PR #28, opened 2014-07-09
WIP: Abstract algorithmic interface for iterative orthogonalization

jiahao:
This PR will eventually provide several iterative orthogonalization algorithms for constructing an orthogonal basis for a vector space. The literature makes a big deal about controlling for loss of orthogonality in iterative methods, and so this is my attempt to throw the kitchen sink at the problem by providing the more widely touted variants of Gram-Schmidt orthogonalization. (If nothing else, users should be able to test these claims of orthogonality loss by having fine-grained control over the orthogonalization steps.)

The key design consideration here is how to provide these methods, which form the underlying basis (so to speak) of iterative linear solvers and eigensystem solvers, in a way that is composable and user-controllable.  The interface exposed herein is inspired by the sorting algorithms in `Base` and the [SortingAlgorithms.jl](http://github.com/JuliaLang/SortingAlgorithms.jl) package, where a single generic function is declared and methods are defined for each concrete algorithm. Similarly, this PR exposes new overloaded methods for `qrfact!` and dispatches on abstract `OrthogonalizationAlg <: Algorithm` subtypes. Unlike the sorting algorithms, however, the `Algorithm` subtypes also wrap parameters controlling for the number of iterations. The hope is that the abstract `Algorithm` interface will compose naturally with higher-level linear solvers and eigensolvers, allowing for the possibility of mixing and matching different orthogonalization routines over the course of an iterative solve.

To-do
- [x] Implement classical Gram-Schmidt
- [ ] Implement modified Gram-Schmidt
- [x] Implement columnwise modified Gram-Schmidt
- [x] Implement the Pythagorean variant of the QR normalization routine, which loses precision more slowly
- [x] Refactor the existing `qrfact!` methods to use the `normalize!` methods
- [ ] Implement the Cholesky variant of Gander, 1980
- [ ] Implement Rutishauser's superothogonalization
- [x] Implement backward- and forward- reorthogonalizing variants
- [ ] Implement adaptively reorthogonalizing variants using Rutishauser's and Giraud-Langou's criteria
- [ ] Implement Businger-Golub style partial QR decompositions (using a Frobenius norm threshold)
- [ ] Implement devectorized methods for dense matrices and vectors
- [ ] Allow the orthogonalization to be continued from a given partially orthogonalized iteration state
- [ ] Tests using Lauchli's example

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,27 @@ This package provides iterative algorithms for solving linear systems, eigensyst
 - [Issue #1](https://github.com/JuliaLang/IterativeSolvers.jl/issues/1) documents the implementation roadmap.
 - The interfaces between the various algorithms are still in flux, but aim to be consistent.
 
+## Orthogonalization methods for `qrfact!`
+
+This package provides new methods for `qrfact!` and uses an abstract algorithm interface that inherits from `Base.Algorithm`. The method signature looks like
+
+    qrfact!(A, alg(params...))
+
+Valid values for `alg` are:
+
+- `cgs`: classical Gram-Schmidt
+- `cmgs`: columnwise modified Gram-Schmidt
+
+### Normalization algorithms
+
+    normalize!(v, alg)
+
+Valid values for `alg` are:
+
+- `norm_none`: no normalization
+- `norm_naive`: naïve normalization using `norm()` (default)
+- `norm_pythag`: Pythagorean normalization ([doi:10.1007/s00211-006-0042-1](http://dx.doi.org/10.1007/s00211-006-0042-1)) - takes as parameter an `AbstractVector` which represent a column of previously computed `R` factors.
+
 ## Consistent interface
 
 ### Nomenclature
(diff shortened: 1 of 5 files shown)


haampie:
I think not much is used in iterative methods besides classical & modified Gram-Schmidt, as you're expanding with one vector each time. This might be different in block methods, but we don't yet have them. Also, when it comes to Ax = b and GMRES: nothing fancier than modified Gram-Schmidt is necessary, because finding the solution and losing orthogonality happens at the same time. In thick restarted variants this is different as you know.

When it comes to eigenvalue problems, then it's important to have stable methods. And if I'm not mistaken applying classical Gram-Schmidt twice is Good Enough™. Also: classical Gram-Schmidt twice in BLAS-2 form is as fast as modified Gram-Schmidt in BLAS-1 form on modern architectures.

At the moment I've already added the iterative variant of classical Gram-Schmidt [here](https://github.com/JuliaMath/IterativeSolvers.jl/blob/81e78728631ed610953c6d0869cd717ab7c8f7df/src/orthogonalize.jl) and I doubt we need more than that.

haampie:
@jiahao, what can we best do with this PR? See also my previous comment. 

It seems like a lot of hard work has gone into this!

haampie:

PR #29, opened 2014-07-18
WIP: Experimental version of cg_hs iterator

jiahao:
This branch contains an implementation of conjugate gradients as an iterator. The algorithm is implemented entirely using the Julia iterator protocol, which implements methods for `start`, `next` and `done`.

In this paradigm:
- `start` initializes the algorithm
- `next` takes a single conjugate gradient step, and 
- `done` checks the termination criteria

This branch introduce new types:
- `IterativeSolver` encapsulates the parameters used to initialize the iterative algorithm, which is further decomposable into:
  - `KrylovSpace`: the associated Krylov subspace
  - `Terminator`: all the termination criteria, which is composed of
    - `stopcriterion`: a single termination criterion
- `IterationState` represents the state of the iterative algorithm

Conceptual todos:
- [ ] Figure out a representation of preconditioners. I think they are associated with the `KrylovSpace`...
- [ ] Figure out a representation of `ConvergenceHistory` in this design space

Thoughts so far:
- The separation of iteration state has helped clarify how cg_hs works; there is a clear separation of old and new quantities.
- The iterator protocol allows for a very clean abstraction of termination criteria. Just about every iterative solver algorithm here will be able to use the same `Terminator` and `stopcriterion` pieces to test for `done`ness.
- There is also a cute trick here (which may be too cute) in which a relative norm criterion is initializable even when the initial norm is not known, and is swapped out in favor of an absolute norm criterion once such the norm of the residual vector is computed.

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -1,7 +1,19 @@
+#######################
+# Conjugate gradients #
+#######################
+
 export cg, cg!
 
 import Base: start, next, done
+
+@doc doc"""
+Abstract iterative solver
+""" ->
 abstract IterativeSolver
+
+@doc doc"""
+Abstract state of iterative solver
+""" ->
 abstract IterationState
 
 immutable KrylovSpace #XXX to replace KrylovSubspace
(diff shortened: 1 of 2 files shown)


haampie:
See also #137, that is a slightly simpler implementation, but "advanced" users can basically override the stopping criterion. Should this PR continue to be open once #137 is merged?

andreasnoack:
Superseded by #137

PR #31, opened 2014-09-12
WIP: Implicitly restarted Arnoldi factorizers

jiahao:
This PR contains very naïve (but working!) implementations of the Arnoldi method for iteratively constructing partial Hessenberg factorizations (equivalently, Arnoldi factorizations, incomplete Hessenberg factorizations, or truncated Hessenberg factorizations), as encapsulated in the `ArnoldiFactorization` type which wraps the partial Hessenberg factorization (_H_, _V_) and the residual vector _r_ such that

image

Explicitly restarted and implicitly restarted variants of Arnoldi are also implemented as `Restarted{Arnoldi}` and `ImplicitlyRestarted{Arnoldi}`. (The implicitly restarted Arnoldi method is the core numerical algorithm underlying ARPACK.) These iterators are constructed from `Restarted` and `ImplicitlyRestarted`; both are parametric `Factorizer` types which are themselves parametric on an underlying `Factorizer`. The idea is to factor out the the techniques for restarting the factorization process from the underlying factorization process itself. (It could easily be symmetric Lanczos or biorthogonal Lanczos instead, where the underlying matrix factorization object is the tridiagonal analogue of the Arnoldi factorization.)

There is a lot to do here:
- [ ] Figure out what to export
- [ ] The `Arnoldi` process should become parametric on the orthogonalization method (#28)
- [ ] Consolidate the `Arnoldi` iterator with implementations in other algorithms like GMRES
- [ ] The `Terminator` type should be consistent with #29
- [ ] The `Shifts` function should provide most (if not all) of the various spectral partitioning methods described in the ARPACK documentation. Currently, only the `LM` (largest magnitude) mode is implemented. The nice thing is that Julia allows users to specify a custom rule for partitioning the spectrum as a function (the `by` keyword argument of `Shifts`).
- [ ] The deflation step in `ImplicitlyRestarted{Arnoldi}` should be replaced with the bulge-chasing version of the implicitly shift QR update, which is much more efficient.
- [ ] Is there a "thin" version of the matrix factorization? What is the bare minimum state needed to be saved during the iteration process?

### DIFF
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -184,6 +184,38 @@ for i=1:5, j=1:min(5,i+1); H[j,i]=randn(); end
 println(round(H,3))
 Shifts(H, 3)
 
+
+# Get approximate eigenvalues and eigenvectors
+# The usual method is to approximate them with the Ritz vectors and values
+# satisfying the Galerkin condition
+#
+#     w⋅(Ax - xθ) = 0 ∀w ∈Krylov(A,v) of size k
+#
+# The Ritz values are just the eigenvalues of Ar.H
+# and the Ritz vectors can be formed by lifting the eigenvectors of Ar.H
+# back into the original basis (essentially undoing the projection into the
+# Krylov subspace).
+#
+# Par80 gives additional rigorous error bounds on the eigenvalues.
+#
+# purge=false (default) performs an additional correction to the eigenvectors
+# known as purging as suggested in Ericsson and Ruhe, 1980.
+# Use purge only if the computed eigenvalues are suspected to be large
+# (purge=true recommended for shift-and-invert transformed problems)
+function eigfact!(Ar::ArnoldiFact; purge::Bool=false)
+    fact = eigfact!(Ar.H)
+    evals, evecs = fact.values, Ar.V*fact.vectors
+    if purge
+        for i in 1:length(evals)
+            evecs[:,i] += fact.vectors[end,i]/evals[i]*Ar.r
+            evecs[:,i] /= norm(evecs[:,i])
+        end
+    end
+    Base.Eigen(evals, evecs)
+end
+
+eigvals(Ar::ArnoldiFact) = eigvals(Ar.H)
+
 # Test of implicitly and explicitly restarted Arnoldi iterators
 n=15
 M=randn(n,n)+im*randn(n,n)
@@ -200,12 +232,6 @@ for (iter, Ar) in enumerate(Restarted(Arnoldi(K, Terminator(1e-9, n)), n))
     @assert abs(norm(K.A*Ar.V-Ar.V*Ar.H) - norm(Ar.r)) < sqrt(eps())
 end
 
-function eigfact!(Ar::ArnoldiFact)
-    fact = eigfact!(Ar.H)
-    fact.values, Ar.V*fact.Q
-end
-
-eigvals(Ar::ArnoldiFact) = eigvals(Ar.H)
 
 # # References
 #
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -190,6 +190,11 @@ for (iter, Ar) in enumerate(Restarted(Arnoldi(K, Terminator(1e-9, n)), n))
     @assert abs(norm(K.A*Ar.V-Ar.V*Ar.H) - norm(Ar.r)) < sqrt(eps())
 end
 
+function eigfact!(Ar::ArnoldiFact)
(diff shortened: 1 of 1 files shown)


yixuan:
This is amazing!

I jumped to this page from Tony Kelman's slides about Julia's dependencies, and just noticed that more than one year ago you already had a Julia implementation of the Arnoldi method.

FYI, this summer I tried to write a C++ library that can be used as an alternative to ARPACK with a clearer structure and API, and here is the repository: https://github.com/yixuan/spectra. It depends on [Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page), but both Eigen and the library itself are header-only, so it can be easily embedded into other projects.

I'm not sure if this is helpful for you since it seems that a pure Julia implementation is preferred. But at least it provides a prototype of the algorithms that is much easier to read than the FORTRAN code in ARPACK. Here I also have some benchmark results of it compared with ARPACK, http://yixuan.cos.name/spectra/performance.html, and it seems to match ARPACK well.

ViralBShah:
Off topic - @yixuan we hope to soon get to a point where you can write these algorithms in Julia and compile them to a C/Fortran callable shared library.

yixuan:
@ViralBShah OK. Let me write some sample code first and then try to port it to Julia.

yixuan:
Hi @jiahao @ViralBShah 

I've created a repository to implement the symmetric eigensolver in Julia: https://github.com/yixuan/Spectra.jl.

From the documentation of IterativeSolvers.jl it seems that the interface is mainly for linear solvers and it is unclear what would be the best API for eigensolvers, so I just mimic `eigs()` and created the `symeigs()` function.

I'm pretty much a beginner of Julia, so the code may be sub-optimal. The [test script](https://github.com/yixuan/Spectra.jl/blob/master/test/runtests.jl) contains a tiny benchmark between `symeigs()` and `eigs()`, and below is the result on my machine:

```
Matrix size: 100
Error: 1.5631940186722204e-13
Time of eigs(): 0.015242132
Time of symeigs(): 0.035714849

Matrix size: 1000
Error: 3.410605131648481e-13
Time of eigs(): 1.007082048
Time of symeigs(): 1.15339598
```

@tkelman 's [slides](https://github.com/tkelman/BAJUtalk-Dec2014/blob/master/BAJUtalkDec2014.pdf?raw=true) mentioned that Julia team was trying to replace ARPACK with pure Julia code. I'm not sure if this is still your future plan, but if so, I'm very glad to contribute. :-)

ViralBShah:
We definitely would like to do so. Pinging @andreasnoack as well.

jiahao:
Thanks for your contribution! Yes, we would be interested in having a Lanczos/Arnoldi-based eigensolver.

I stopped work on this branch because it turns out that the thick restart method is mathematically equivalent to implicit restarts and the former is also significantly simpler to implement. The implicit restart in this branch uses brute force QR to do the implicit shifts, which is the slowest way. Thick restart is now available for the SVD based upon Lanczos bidiagonalization - see the current implementation of `svdl`.

jiahao:
@yixuan I had a look at your code in Spectra.jl and it looks like you are using the proper Givens-based QR method and also have essentially the same methods for selecting eigenvalues as in ARPACK. If you'd like to submit the code in Spectra.jl as a PR to this package, we can close this PR.

yixuan:
Thanks for the comments. I'll try to submit a PR later.

yixuan:
Hello @jiahao , sorry for the absence of the PR for such a long time. At that moment I detected a numerical issue that has just been fixed recently, so I will continue from now on.

jiahao:
Thanks for the update! Looking forward to getting a PR

src/arnoldi.jl
```diff
@@ -105,6 +97,98 @@ for (iter, Ar) in enumerate(Arnoldi(K, Terminator(1e-9, n)))
...
+        u = P.getshifts(H, P.p)
+
+        Q = I
+        for j=1:p
+            Qj, Rj = qr(H-u[j]*I)
+            H = Qj'H*Qj #TODO replace with bulge-chasing
+            Q = Q*Qj
+        end
+
+        VQ = V*Q
```
haampie: I think this should be done in-place, because V is most memory-consuming. If you do restarts because of memory limitations, then this procedure would require you to restart twice as early. One way to go is use the LU-trick: V * Q = (V * L) * U. Since L & U are upper triangular, this can be done in-place.

haampie:
@jiahao what is the status of this PR? I know it has been a while ^^

haampie:

PR #40, opened 2015-07-22
Fix tests

andreasnoack:
Added `Compat` dependency for this.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.3
 Docile
-
+Compat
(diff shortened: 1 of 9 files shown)


andreasnoack:
Fixing the tests revealed a bug in Julia when solving symmetric eigenvalue problems for empty matrices. This is now fixed on both master and 0.3, but this will only have effect on the 0.3 tests after next release. I've tested locally on latest 0.3 and the tests pass so I'll go ahead an merge,

jiahao:
+1

PR #46, opened 2015-09-01
Add IDR(s) solver from https://github.com/mschauer/IDRsSolver.jl,

### LINKED ISSUE #39: https://github.com/mschauer/IDRsSolver.jl
@astudillor and me ported a basic Induced Dimension Reduction Solver to julia, @tkelman mentioned IterativeSolvers and medium-term I think this could be incorporated into this package. Anyway I harmonized the interface mostly, modulo issues how generic LinearOperators are treated.

mschauer:
``idrs_core!`` is written to accept matrix shaped ``x`` and ``b`` and general linear operators op(x, ...)=b, but ``idrs`` self follows the  gmres function.

Ping @astudillor.

Closes #39 .

### DIFF
--- a/src/idrs.jl
+++ b/src/idrs.jl
@@ -2,19 +2,20 @@ export idrs, idrs!
 
 ####
 
-function omega(t, s, angle = .7)
+@inline function omega(t, s)
+    angle = sqrt(2.)/2
     ns = vecnorm(s)
     nt = vecnorm(t)
     ts = vecdot(t,s)
     rho = abs(ts/(nt*ns))
     om = ts/(nt*nt)
-    if rho < angle 
-        om = om*angle/rho
+    if rho < angle
+        om = om*convert(typeof(om),angle)/rho
     end
     om
 end
 
-linsys_op(x, A) = A*x
+@inline linsys_op(x, A) = A*x
 
 """
 IDRsSolver.jl
@@ -80,9 +81,9 @@ idrs!(x, A, b; s = 8, tol=sqrt(eps(typeof(real(b[1])))), maxiter=length(x)^2) =
     idrs_core!(x, linsys_op, (A,), b, s, tol, maxiter)
 
 function idrs_core!{T}(X, op, args, C::T, s, tol, maxiter)
-    R = C - op(X, args...)
+    R = C - op(X, args...)::T
     normR = vecnorm(R)
-    res = [normR]
+    res = typeof(tol)[normR]
 	iter = 0                 
 
     if normR <= tol           # Initial guess is a good enough solution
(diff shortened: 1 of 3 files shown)


mschauer:
Could I get some feedback here?

src/idrs.jl
```diff
@@ -0,0 +1,183 @@
...
+    iter = 0
+    while normR > tol && iter < maxiter
+        for i in 1:s,
+            f[i] = vecdot(P[i], R)
+        end
+        for k in 1:s 
+
+            # Solve small system and make v orthogonal to P
+
+            c = \(LowerTriangular(M[k:s,k:s]),f[k:s])
```
andreasnoack: Infix is more readable here

src/idrs.jl
```diff
@@ -0,0 +1,183 @@
...
+            axpy!(beta, U[k], X)
+        
+            normR = vecnorm(R)
+            res = [res; normR]
+            iter += 1
+            if normR < tol || iter > maxiter
+                return X, ConvergenceHistory(normR < tol, tol, length(res), res)
+            end 
+            if k < s 
+                f[k+1:s] = f[k+1:s] - beta*M[k+1:s,k]
```
andreasnoack: `axpy!`?

src/idrs.jl
```diff
@@ -0,0 +1,183 @@
...
+            end
+        
+        end
+
+        # Now we have sufficient vectors in G_j to compute residual in G_j+1
+        # Note: r is already perpendicular to P so v = r
+
+        copy!(V, R)
+        Q = op(V, args...)
+        om = omega(Q, R, 0.7)
```
andreasnoack: `0.7`? Is that a magic number from the papers?

andreasnoack:
Have you tried to use `@code_warntype` on the implementation to see if the compiler can infer the types for e.g. `Float64` and `Float32` inputs?

src/idrs.jl
astudillor: well cos(pi/4) is the magic number, but in the papers they approximate this by 0.7

andreasnoack: Maybe it doesn't really matter, but I think you can get the exact value for free because LLVM will compute the value as a constant
```
julia> f() = sqrt(2)/2;

julia> @code_llvm f()

; Function Attrs: sspreq
define double @julia_f_21376() #0 {
top:
  ret double 0x3FE6A09E667F3BCD
}
```

src/idrs.jl
mschauer: Then I have to take SubArrays, I don't think that makes much difference.

mschauer:
Thanks! I made some tweaks regarding type stability. There are still some Any-warnings which I think are spurious, for example``#s34 = (Base.box)(Base.Int,(Base.add_int)(#s34::Int64,1)::Any)::Int64``.

andreasnoack:
Great. I think this is fine now, but I'd like to hear what is causing the OS X error even though I think it is completely unrelated to this pr.

andreasnoack:
Sorry for the delay. The Mac builds just started to work again.

Issue #33, opened 2015-01-20
Randomized algorithms roadmap


jiahao:
This issue tracks the implementation of randomized algorithms that have appeared in the literature.

Dixon, 1983 [doi:10.1137/0720053](http://epubs.siam.org/doi/abs/10.1137/0720053):
- [x] Randomized condition number estimate (implemented as `rcond`)
- [x] Randomized extremal eigenvalue estimates (implemented as `reigmax` and `reigmin`)

Cheng, Gimbutas, Martinsson and Rokhlin, 2005 [doi:10.1137/030602678](http://epubs.siam.org/doi/abs/10.1137/030602678)
- [ ] Interpolative decomposition (the four-term variant)

Liberty, Woolfe, Martinsson, Rokhlin, Tygert, 2007 [doi:10.1073/pnas.0709640104](http://www.pnas.org/content/104/51/20167)
- [x] Randomized norm estimate (implemented as `rnorms`)
- [x] Interpolative decomposition (the two-term variant; currently a very hacky version is implemented in `idfact`)

Strohmer and Vershynin, 2009 [doi:10.1007/s00041-008-9030-4](http://link.springer.com/article/10.1007%2Fs00041-008-9030-4): 
- [ ] randomized Kaczmarz algorithm

Halko, Martinsson and Tropp 2011 [doi:10.1137/090771806](http://epubs.siam.org/doi/abs/10.1137/090771806)
- [x] Lemma 4.1: Randomized norm estimate (implemented as `rnorm`)
- [x] Algorithm 4.1: randomized range finder (implemented as `rrange`)
- [x] Algorithm 4.2: adaptive randomized range finer (implemented as `rrange_adaptive`)
- [x] Algorithm 4.4: randomized subspace iteration (implemented as `rrange_si`)
- [x] Algorithm 4.5: fast randomized range finder (implemented as `rrange_f`)
- [ ] Variant of Algorithm 4.5 using randomized Givens rotations
- [x] Algorithm 5.1: direct SVD (implemented as `svdfact_restricted`)
- [x] Algorithm 5.2: SVD via row extraction (implemented as `svdfact_re`) 
- [x] Algorithm 5.3: direct eigenvalue decomposition (implemented as `eigfact_restricted`)
- [x] Algorithm 5.4: eigenvalue decomposition via row extraction (implemented as `eigfact_re`) 
[...]

PR #47, opened 2015-09-04
A randomized kaczmarz algorithm

matthieugomez:
Randomized kazmarz algorithm (Strohmer and Vershynin, 2009). On the roadmap #33 

I implemented a version for full and sparse matrices.

Feedback welcome!

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 Compat
+Distributions
 julia 0.4-
 
--- a/src/IterativeSolvers.jl
+++ b/src/IterativeSolvers.jl
@@ -29,5 +29,6 @@ include("lsqr.jl")
 include("rlinalg.jl")
 include("rsvd.jl")
 include("rsvd_fnkz.jl")
+include("rkaczmarz.jl")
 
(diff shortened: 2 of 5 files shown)


mschauer:
You already made same amends, right? The result should be returned together with a 
```
type ConvergenceHistory{T, R}
    isconverged::Bool
    threshold::T
    mvps::Int
    residuals::R
end
```

mschauer:
Not being a collaborator of the project, but I anticipate that the following would be helpful:

Did you test you implementation for different element-types, especially for Complex floats? Currently the tests only covers a real matrices, right? It would be good for the test to follow https://github.com/JuliaLang/IterativeSolvers.jl/blob/master/test/runtests.jl and test on the same kind of problems as say GMRES using the same protocol based on FactCheck.jl.

Then you could also have a look at the memory consumption of the algorithm when n grows for different matrix types so one can see for example if somewhere needless copies of column vectors are made.

haampie:
@matthieugomez, can we still try to get the Kaczmarz method into this package?

PR #51, opened 2015-09-26
lsmr

matthieugomez:
implementation of the lsmr method (see issue [#3531 in METADATA.jl](https://github.com/JuliaLang/METADATA.jl/pull/3531))
I'd like to use this implementation in a code that needs to be efficient, and so there are two differences with lsqr:
- a supplementary signature that allows to pass preallocated arrays (ensuring no array is created within the function)
- Using `A_mul_B!(alpha, A, x, beta, y)` transformations rather than `A_mul_B!(y, A, x)` 

Similar changes could be done in `lsqr`, but I did not touch it.

### DIFF
--- a/src/lsmr.jl
+++ b/src/lsmr.jl
@@ -1,0 +1,247 @@
+export lsmr, lsmr!
+
+using Base.LinAlg 
+
+##############################################################################
+## LSMR
+##
+## Minimize ||Ax-b||^2 + λ^2 ||x||^2
+##
+## Adapted from the BSD-licensed Matlab implementation at
+## http://web.stanford.edu/group/SOL/software/lsmr/
+##
+## A is a StridedVecOrMat or anything that implements 
+## A_mul_B!(α, A, b, β, c) updates c -> α Ab + βc
+## Ac_mul_B!(α, A, b, β, c) updates c -> α A'b + βc
+## eltype(A)
+## size(A)
+## (this includes SparseMatrixCSC)
+## x, v, h, hbar are AbstractVectors or anything that implements
+## norm(x)
+## copy!(x1, x2)
+## scale!(x, α)
+## axpy!(α, x1, x2)
+## similar(x, T)
+## length(x)
+## b is an AbstractVector or anything that implements
+## eltype(b)
+## norm(b)
+## copy!(x1, x2)
+## fill!(b, α)
+## scale!(b, α)
+## similar(b, T)
+## length(b)
+
+##############################################################################
+
+
+## Arguments:
+## x is initial x0. Transformed in place to the solution.
+## b equals initial b. Transformed in place
+## v, h, hbar are storage arrays of length size(A, 2)
+function lsmr!(x, A, b, v, h, hbar; 
+    atol::Number = 1e-6, btol::Number = 1e-6, conlim::Number = 1e8, 
+    maxiter::Integer = max(size(A,1), size(A,2)), λ::Number = 0)
+
+    # Sanity-checking
+    m = size(A, 1)
+    n = size(A, 2)
+    length(x) == n || error("x has length $(length(x)) but should have length $n")
+    length(v) == n || error("v has length $(length(v)) but should have length $n")
+    length(h) == n || error("h has length $(length(h)) but should have length $n")
+    length(hbar) == n || error("hbar has length $(length(hbar)) but should have length $n")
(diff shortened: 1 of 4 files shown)


andreasnoack:
Looks good to me. Thanks for checking the code for type instabilities. Have you looked at it with `@code_warntype`?

I think your two bullets are good choices. I've wanted to change the multiplication signature in base to `A_mul_B!(α, A, x, β, υ)`, but haven't done it yet.

matthieugomez:
I mainly copied the code in `lsqr` for type stability. Although `@code_warntype` tells me that the return value is correctly typed, I could not use it to examine what happens in the function body itself.

```julia
using IterativeSolvers
x = zeros(10)
A = rand(10, 10)
b, v, h, hbar = rand(10), rand(10), rand(10), rand(10)
@code_warntype lsmr!(x, A, b, v, h, hbar)
Variables:
  x::Array{Float64,1}
  A::Array{Float64,2}
  b::Array{Float64,1}
  v::Array{Float64,1}
  h::Array{Float64,1}
  hbar::Array{Float64,1}

Body:
  begin $(Expr(:line, 28, symbol("/Users/Matthieu/.julia/v0.4/IterativeSolvers/src/lsmr.jl"), symbol("")))
      GenSym(1) = (Base.arraysize)(A::Array{Float64,2},1)::Int64
      GenSym(0) = (Base.arraysize)(A::Array{Float64,2},2)::Int64
      return (IterativeSolvers.__lsmr!#54__)(1.0e-6,1.0e-6,1.0e8,((top(getfield))(Base.Intrinsics,:select_value)::I)((Base.slt_int)(GenSym(0),GenSym(1))::Bool,GenSym(1),GenSym(0))::Int64,0,x::Array{Float64,1},A::Array{Float64,2},b::Array{Float64,1},v::Array{Float64,1},h::Array{Float64,1},hbar::Array{Float64,1})::Tuple{Array{Float64,1},IterativeSolvers.ConvergenceHistory{Tuple{Float64,Float64,Float64},Array{Tuple{Float64,Float64,Float64},1}}}
  end::Tuple{Array{Float64,1},IterativeSolvers.ConvergenceHistory{Tuple{Float64,Float64,Float64},Array{Tuple{Float64,Float64,Float64},1}}}
```

andreasnoack:
Then it is time to learn a trick. The reason you don't see the body of the function is the keywords. Because of them, the function with the actual code is kind of hidden, but you can see the name of it right after the `return` statement. It is defined such that the keyword arguments come first. Therefore you can inspect the body with 
```jl
@code_warntype IterativeSolvers.(symbol("__lsmr!#54__"))(1e-6, 1e-6, 1e8, size(A,1), 0, x, A, b, v, h, hbar)
```
I did it and it looks fine so unless @jiahao has something to add, I'll merge this pr soon.

matthieugomez:
Thanks. After that I'll email the Michael Saunders so that links to LSQR and LSMR in Julia on the [SOL  website](http://web.stanford.edu/group/SOL/software) point to this directory.

matthieugomez:
One thing I'm not sure about is what to return in ch.residuals. Returning `ssr` looses a lot of information wrt quantities computed in `lsmr`. So I return the test corresponding to the tolerance `atol, btol, ctol`, which does not correspond to `ssr`.

src/lsmr.jl
```diff
@@ -0,0 +1,250 @@
...
+			β = norm(u)
+			if β > 0
+				scale!(u, inv(β))
+				Ac_mul_B!(1, A, u, -β, v)
+				α = norm(v)
+				α > 0 && scale!(v, inv(α))
+			end
+		
+			# Construct rotation Qhat_{k,2k+1}.
+			αhat = sqrt(abs2(αbar) + abs2(λ))
```
andreasnoack: Can we be sure that this quantity won't overflow, i.e. that `hypot` is not necessary?

andreasnoack:
I'm not sure we are that consistent about the `residuals` field in the convergence history. The name `residuals` is also a bit misleading if we are storing the residual norms. Therefore, I think it is okay to keep the congence tests for now, but eventually it would be good to be consistent if possible.

You have tabs in the source and our convention is to use four spaces to I'll ask you to change that as well.

matthieugomez:
Updated. Thanks!

andreasnoack:
Thanks for the contribution.

PR #52, opened 2015-10-01
Avert method-overwritten warnings

timholy:


andreasnoack:
The usual SSL error on nightly.

PR #58, opened 2015-11-10
Fix matvecs computation in gmres!

mschauer:
Otherways, the number is given in terms of the next multiple of restart, which is meaningless when maxiter is 1.


jiahao:
Thanks!

PR #61, opened 2015-12-18
modified gmres to avoid an extra application of the linear operator

Forgotten:
I made a small modification to the gmres function in order to avoid a second application of the linear operator at each iteration.


jiahao:
Thanks!

PR #67, opened 2016-02-26
Eigenvalue computation using implicitly restarted Arnoldi method

yixuan:
This is a follow-up of the discussion in #31.

It provides the function `eigvals_arnoldi()` that implements most functionality of `eigs()` on symmetric matrices. I haven't added the documentation yet, since the interface and arguments may change a lot depending on the discussion here. And since this function is intended to replace `eigs()` in the future, the documentation can be easily derived from `eigs()`.

There are some issues here:
1. The interface of this function is not consistent with `eigvals_lanczos()`, due to the nature of the Arnoldi method; in fact it is more similar to `eigs()`.
2. Computed eigenvalues are in decreasing order, to be consistent with `eigs()`.
3. The function is for symmetric linear operator only. I can also try to write a general eigen solver, but it takes more effort.
4. The test code causes one `svdl` test to fail, since it changed the state of the random seed. Sorry...

### DIFF
--- a/src/IterativeSolvers.jl
+++ b/src/IterativeSolvers.jl
@@ -4,6 +4,7 @@ isdefined(:__precompile__) && __precompile__(true)
 
 include("common.jl")
 include("krylov.jl")
+include("hessenqr.jl")
 
 #Specialized factorizations
 include("factorization.jl")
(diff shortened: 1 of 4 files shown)


haampie:
@yixuan what is the status of this PR? Are there plans to getting this merged?

haampie:

PR #68, opened 2016-03-21
Fix Rayleigh quotient iteration

lopezm94:
Hello, I'm a student interested in working on this package's GSoC project.

I've modified Rayleigh quotient iteration method to be the same as the already implemented Inverse method, except at every iteration it approximates the eigenvalue too with the Rayleigh quotient.
The test is a copy of the Inverse's test as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,4 +106,4 @@ Orthogonalizing the basis vectors for a `KrylovSubspace` is crucial for numerica
 - `method`: Orthogonalization method. Currently supported methods are:
   - `:GramSchmidt`. Not recommended, as it is numerically unstable.
   - `:ModifiedGramSchmidt`. (default)
-  - `:Householder`. This is actually the same as `:ModifiedGramSchmidt`.
+  - `:Householder`.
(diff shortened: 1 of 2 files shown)


kshyatt:
I'm bumping this @jiahao @andreasnoack

src/simple.jl
```diff
@@ -56,23 +56,25 @@ function eigvals_ii(A, σ::Number, x=nothing; tol::Real=eps(), maxiter::Int=size
...
+	σ = σ+1/θ
+        resnorms[iter] = norm(y - θ*v)
+        if resnorms[iter] <= tol * abs(θ)
             resnorms=resnorms[1:iter]
             break
         end
+        appendunit!(K, y)
     end
-    Eigenpair(ρ, v), ConvergenceHistory(0<resnorms[end]<tol, tol, K.mvps, resnorms)
+    Eigenpair(σ, y/θ), ConvergenceHistory(0<resnorms[end]<tol*abs(θ), tol, K.mvps, resnorms)
```
jiahao: Is `y/θ` guaranteed to be normalized? In your change, `θ = v ⋅ y` now

lopezm94: When v is equal to y, but that's not always the case if it doesn't converge.

Issue #75, opened 2016-07-09
New API name conventions


lopezm94:
Due to the coming of the new API, the names of some methods and keyword arguments are going to change, these are all preliminary so feel free to share alternatives. Here are some of the changes:

##### Method names:

* eigvals_power => powm
* eigvals_ii => invpowm
* eigvals_rqi => rqi
* gauss_seidel => gs
* eigvals_lanczos => eiglancz

##### Keyword arguments:

In the new API all the optional positional arguments are now going to be keyword arguments. Some of the new keyword names:

* shift: For inverse iteration and rayleigh
* nsv: For number of singular values requested in svdl
* The rest have the same name but are now keywords

##### Dictionary keys:

It is also worth discussing the name conventions in [ConvergenceHistory](https://github.com/JuliaLang/IterativeSolvers.jl/issues/6#issuecomment-231146863) since it will most likely have an inner dictionary for tolerances and the rest of the information. So some names for tolerances could be reltol, abstol and so on instead of atol, btol, ctol or other non-explicit naming, it goes the same for norms (resnorm, Anorm, etc). If two of them repeat then it will have a number as suffix.

@lruthotto @dpo

ChrisRackauckas:
I highly prefer more descriptive names to shortened names which require knowing package details to understand. I.e. `eigvals_lanczos` >> `eiglancz`, though `eig_lanczos` would suffice. `eiglancz` would be a ?? if I didn't already know what it was, and I couldn't even just google the method really quickly. Same with `nsv`. If I see that in code, I don't necessarily know what that is, but `num_singular_val` is readable. I fear that some of this is a little too far on the terse side.

lopezm94:
What I tried to do was put short names on every keyword and solve any misunderstanding in the docs. There you can check anything you want. Is it to messy?

For example:

```
svdl(A)

  Compute some singular values (and optionally vectors) using
  Golub-Kahan-Lanczos bidiagonalization cite{Golub1965} with thick restarting
  cite{Wu2000}.

  If log is set to true is given, method will output a tuple X, L, ch. Where
  ch is a ConvergenceHistory object. Otherwise it will only return X, L.

  The plot attribute can only be used when log is set version.

  Arguents

    •    A : The matrix or matrix-like object whose singular values are
        desired.
      

  Keywords

    •    nsv::Int = 6: number of singular values requested.
      
    •    v0 = random unit vector: starting guess vector in the domain of A.
      

[...]

Issue #60, opened 2015-11-17
GMRES docs need to explain macro- and microiterations better


dprn:
According to the docs, for every solver:
- `maxiter` is the maximum number of allowed iterations, which conventionally defaults to `length(b)`

However this is not true (at least for `gmres`). Indeed, the default number of iterations is 1, which obviously very rarely converge (from https://github.com/JuliaLang/IterativeSolvers.jl/blob/6ab23802336ec7883965c93c6ff23ca33a735e29/src/gmres.jl):

```
gmres(A, b, Pl=1, Pr=1;
      tol=sqrt(eps(typeof(real(b[1])))), maxiter::Int=1, restart::Int=min(20,length(b))) =
    gmres!(zerox(A,b), A, b, Pl, Pr; tol=tol, maxiter=maxiter, restart=restart)
```

mschauer:
Not quite, but I agree it is a bit misleading. ``gmres`` wraps both the restarted and the not-restarted algorithm, ``restart`` gives the number of steps in the inner loop (restarted after), ``maxiter`` the number of outer loops. As ``restart=min(20,length(b))`` and ``maxiter=1`` the default arguments correspond to a not restarted gmres which runs for a total of ``restart*maxiter==min(20,length(b))`` iterations.

mschauer:
With that, do you have an idea for better naming of the parameters?

jiahao:
The standard terms AFAICT are macroiteration and  and microiteration. There is clearly a documentation issue at hand.

zimoun:
From my knowledge, `maxiter` and `restart` are the "classical" names for these parameters. And still from my knowledge, we talk about inner-loop and outer-loop.

Should  modify `restart: Maximum number of inner iterations (GMRES(restart))`
And should add that the total number of iterations is therefore: `maxiter*restart`, and specify that the full `gmres` is `maxiter=1`.

I miss a point: the doc is it the commentary in the top of the function ? why is it not a docstring ?

lopezm94:
For the new GMRES API we are thinking of defining maxiter as total number of iterations, i.e., ((outer-1)*inner + inner_left). The method will then calculate the number of outer iterations as ceil(maxiter/restart)

PR #77, opened 2016-07-28
Api proposal

lopezm94:
## Interface

Here is the API proposal, it contains the name changes stated on issue #75 and also the GMRES doc fix requested on issue #60. The current test and benchmarks are updated for the new syntax.

### Input

Positional arguments not necessary for a method call are now keyword arguments. All methods have two new keyword arguments and two return options:

*Keywords*:

- verbose
- plot

*Return options*:

1. Standard result of the function
2. Standard result of the function, `ConvergenceHistory` (This was the only option before)

For the first one you only need to call as it is currently implemented in the package.
For the second one the method call should always look like this:
```julia
method(Master, ...) #Master is a custom type made only for dispatching
```

### Output

The new `ConvergenceHistory` internals are much different than before:
```julia
type ConvergenceHistory{T}
    mvps::Int    # Matvec products
    mtvps::Int     # Transposed matvec products
    iters::Int     # Total number of iterations
    restart::T    # Restart size
    isconverged::Bool    # Convergence flag
    tol::Dict{Symbol, Real}    # Tolerances
    data::Dict{Symbol, VecOrMat}    # Convergence data (Not necessarily residuals)
end
typealias PlainHistory ConvergenceHistory{Void}
typealias RestartedHistory ConvergenceHistory{Int}
```

Most are self explanatory. The `data` attribute can only have info meant to be stored and updated at each iteration (that's why it only accepts vectors or matrices). For methods like svdl storing information such as the ritz or beta values is now possible inside the `data` attribute.

#### Plots 

For quick and dirty plotting, just set `plot` keyword and info about the method will be plotted using UnicodePlots.

For more advanced  request, `ConvergenceHistory` has custom recipe so it is compatible with [Plots.jl](https://github.com/tbreloff/Plots.jl).

### Docs

As soon as this is merged I'll proceed to complete the docs, for a quick look of the current theme check [here](http://juanlopezm.me/IterativeSolvers.jl/library/public/).


ChrisRackauckas:
Instead of a `showplot`, you should do a plot recipe (to make a `plot(::ConvergenceHistory)` ) to Plots.jl so that it's more customizable and so that you don't have to have a plotting package as a dependency (just RecipesBase).

lopezm94:
Could there be a single recipe for multiple plots? `data::Dict{Symbol, VecOrMat}` will have multiple plot options.

ChrisRackauckas:
Yes, I think you could do a series recipe to handle that. @tbreloff would know.

tbreloff:
Sure that would be straightforward I think. Look at doing a "user recipe"

lopezm94:
Plots.jl is amazing, thanks for the package. If I understand correctly the way to print everything would be to design a vertical `@layout` and do a `@series` for each plot. However, I think I'll make a new plot(::ConvergenceHistory) definition and then a `@userplot` to plot each data element.

lopezm94:
Ok, I added support for Plots.jl via recipes. Do you treat the data in any special way before plotting @tbreloff? whenever I try to to manually plot an error pop ups, but if I use Plots.jl then everything works fine. I ask because we will most likely keep UnicodePlots for easy and quick plotting.

tbreloff:
I peeked at the recipe and it seems like that should work. What error are
you seeing?

lopezm94:
The recipe works perfectly, my question is a bit unrelated to recipes.

Inside the package we do our own plots besides recipes, so since Plots.jl is too big we use `UnicodePlots`, however sometimes this error pops up:

```julia
ERROR: InexactError()
 in Plot at /home/juan/.julia/v0.4/UnicodePlots/src/plot.jl:150
```

But when using your package everything works fine. I was wondering if you process the numbers in any special way before sending them to `UnicodePlots`, so I can emulate it here.

tbreloff:
I see. Can you convert your other plotting to recipes as well?

ChrisRackauckas:
>  so since Plots.jl is too big we use UnicodePlots

You don't need to have Plots.jl as a dependency for defining recipes, just `RecipesBase` (which is rather small).

lopezm94:
I think I'm not expressing myself correctly. I know `RecipesBase` is lightweight, but if I want to plot something without having the user import `Plots.jl` (if he wants to plot the recipe thats up to him), then this package needs another dependency since `RecipesBase` can't plot on it s own. That's why currently this PR plots with `UnicodePlots`. For example:

```julia
svdl(Master, rand(10,10), plot=true)  #Supposed to plot on its own
```

So in this call `IterativeSolvers.jl` needs to plot, recipes can't be plotted without Plots.jl and that's why I use `UnicodePlots` instead. What I was trying to say is that when I call `UnicodePlots` on my own (no recipes, no Plots.jl ) I get this inexact errors sometimes, that's all in I'm saying. Was just hoping tbreloff would know because his package needs to deal with `UnicodePlots` API and issues. I imagine it must be because of some NaN number that I'm passing.

ChrisRackauckas:
I see what you mean.

lopezm94:
It is a `UnicodePlots` bug (see Evizero/UnicodePlots.jl/issues/51). I'll try to make a PR

lopezm94:
The [docs](http://juanlopezm.me/IterativeSolvers.jl/) are out if anyone is interested.

lopezm94:
This PR became too big, I'll separate this in smaller PRs. Sorry for the troubles and the relentless coveralls messages.

PR #80, opened 2016-08-17
Turn off @codecov comments again (sigh)

jiahao:
Codecov has redone their API and reset everything to the default behavior...


KristofferC:
LOL!

PR #81, opened 2016-08-17
Fix to work on Julia 0.5

jiahao:
`sub(M, n, :)` no longer returns a matrix; use `sub(M, n:n, :)` instead.

Also relax test tolerance in `test/svdl.jl`


lopezm94:
In #77 there is a patch for this, using view in v.0.5 and sub in v.0.4

tkelman:
Or use `Compat.view`, much better for bisecting than an imprecise VERSION cutoff.

jiahao:
Closing in favor of #82

PR #82, opened 2016-08-18
Add view instead of slice and sub

lopezm94:
Move to 0.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: julia
 os:
     - linux
 julia:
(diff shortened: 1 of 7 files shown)


jiahao:
Looks good. The only thing that needs to be added is a change to `REQUIRE` to go from `julia 0.4` to `julia 0.5-`. It also looks like we can remove the dependency on `Compat` since it is not used anywhere.

tkelman:
This is in Compat, you don't need to drop 0.4 support for this...

PR #84, opened 2016-08-18
Add dispatch to multiplication and division on Type{Function}

lopezm94:
Remove some unnecessary checks and allow free use of functions.

### DIFF
--- a/src/rlinalg.jl
+++ b/src/rlinalg.jl
@@ -299,6 +299,7 @@ Reference:
 
     \[Equation 4.6]{Halko2011}
 """ ->
+#Define two methods here to avoid method ambiguity with f::Function*b::Any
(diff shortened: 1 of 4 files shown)


src/rlinalg.jl
```diff
@@ -299,10 +299,9 @@ Reference:
 
     \[Equation 4.6]{Halko2011}
 """ ->
-function *(A, Ω::srft)
+*(A::Function, Ω::srft) = function *(A, Ω::srft)
```
jiahao: It would be worth adding a comment along the lines of

```
#Define two methods here to avoid method ambiguity with f::Function*b::Any
```

tkelman:
I think this should be reverted, this is "type piracy" since these types and methods are all owned by Base, this changes the behavior of other unrelated code when IterativeSolvers is imported.

PR #85, opened 2016-08-18
Replace AbstractMatrixCFcn for new type LinearMap

lopezm94:
Replace `AbstractMatrixCFcn` and children for a new unique type `LinearMap`. LinearMap can be created without any functions and defaults them to identity, multiplication and ctranspose multiplication can be added individually. It also switches the functions when it is ctransposed to know at every moment which function to use.

Simpler to use and there is only one type for the user to handle.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -53,13 +53,13 @@ type PosSemidefException <: Exception
     PosSemidefException(msg::AbstractString="Matrix was not positive semidefinite") = new(msg)
 end
 
-export FuncMatrix
+export LinearMap
 
 """
-    FuncMatrix{T,A,B}
+    LinearMap{T,A,B}
 
 Represent functions as a matrix. `T` is the type of the elements inside of the
-'matrix' `(T == eltype(::FuncMatrix))`.
+'matrix' `(T == eltype(::LinearMap))`.
 
 **Fields**
 
@@ -70,8 +70,8 @@ Represent functions as a matrix. `T` is the type of the elements inside of the
 
 **Constructors**
 
-    FuncMatrix(A)
-    FuncMatrix(
+    LinearMap(A)
+    LinearMap(
         m::Int, n::Int; typ::Type=Float64, ctrans::Bool=false,
         mul::Function=identity, cmul=identity
     )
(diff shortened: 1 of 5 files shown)


andreasnoack:
Maybe you are aware but remember that functions have types in 0.5 so it might not be necessary to wrap them in a type. E.g. you can define

```julia
Base.eltype(::typeof(f)) = Float64
Base.size(::typeof(f)) = (10,10)
Base.*(::typeof(f), x) = f(x)
```

I'm not sure if it fits all your needs here but I could get a long way with this recently when I wrote an iterative algorithm.

lopezm94:
I think it won't help in this case, the type wraps two functions: multiplication and ctranspose multiplication. When ctransposing it is desirable to have the wrapper to organize which function to use.

jiahao:
In the code it would be worth adding comments as to what the `A` and `B` type parameters mean.

jiahao:
@lopezm94 has tried the function type approach in this branch: https://github.com/lopezm94/IterativeSolvers.jl/blob/FunctionDispatch/src/common.jl#L79-L117

Basically the main problems are that of scope and delaying the evaluation of the definitions.

@andreasnoack any thoughts?

andreasnoack:
Can either of you explain the "delaying" part to me?

Couldn't it be e.g. `Base.eltype(::typeof(mu)) = typ` instead or am I missing something?

lopezm94:
Reason:

```julia
julia> function f(g::Function)
           g(::typeof(g)) = nothing
       end
ERROR: syntax: cannot add method to function argument g
```

Julia does something `::typeof` at compile time, so there is a need to delay the evaluation

```julia
julia> function f(g::Function)
           @eval g(::typeof(g)) = nothing
       end
f (generic function with 1 method)
```

lopezm94:
Reason:

```julia
julia> function f(g::Function)
           k(::typeof(g)) = nothing
       end
ERROR: syntax: cannot add method to function argument g
```

Julia does something to `::typeof` at compile time, so there is a need to delay the evaluation or it will go out of scope

```julia
julia> function f(g::Function)
           @eval k(::typeof($g)) = nothing
       end
f (generic function with 1 method)

julia> f(identity)
k (generic function with 1 method)
```

lopezm94:
And then there's the problem with the scope, if the user deletes the object the functions might be left behind.

lopezm94:
Closing in favor of #104

PR #94, opened 2016-08-27
Add recipe for ConvergenceHistory and some tests

lopezm94:
Add two recipes for `ConvergenceHistory`, `plot(::ConvergenceHistory)` will plot every field that is a vector or matrix. Vector will be plotted as a series and Matrices as scatterplots. If there is no plotable data , nothing will be done.

There's a second recipe `plot(::ConvergenceHistory,key::Symbol)` to plot a single element inside the data fieldname. If the element isn't a vector or matrix an error will pop up.

If the method has restarts a vertical separator will be added between each restart.

The tests added only check the plots don't crash. Currently `UnicodePlots.jl` outputs some UTF8 warnings on julia 0.5.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.5-
 
 UnicodePlots
+RecipesBase
--- a/src/history.jl
+++ b/src/history.jl
@@ -308,3 +308,66 @@ function plot_collection{T<:Real}(vals::Matrix{T}, iters::Int, gap::Int;
     end
     plot
 end
+
+## Recipes (See Plots.jl tutorial on recipes)
+
+using RecipesBase
+
+# Plot entire ConvergenceHistory. `sep` is the color of the restart separator.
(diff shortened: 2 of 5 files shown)


lopezm94:
(*Before API Change and using GR backend*)

If using UnicodePlots.jl backend the separator will have a random color because of how Plots.jl interacts with it, other backends don't have this problem. Default separator is `:white`.

Example:

```julia
julia> using IterativeSolvers

julia> using Plots

julia> A = rand(100,100); b = rand(100);

julia> hist = gmres(rand(1000,1000),rand(1000), maxiter=10, restart=10)[2];

julia> Pkg.add("GR")

julia> gr();

julia> plot(hist, sep= :black)
```
Output:

screenshot from 2016-09-02 22-56-15

Another example:

```julia
julia> A = rand(100,100); b = rand(100);

julia> hist = cg(rand(1000,1000),rand(1000))[2];

julia> plot(hist)
```

Output:

cg

PR #97, opened 2016-11-29
Shrink history when logging in addition to when plotting

tkonolige:
This patches correctly truncates convergence history so that it only contains valid values. Previously, history was only truncated when plotting (and not when logging).


lopezm94:
Thank you, totally missed it.

PR #104, opened 2017-01-07
WIP: Attempting to integrate linear maps in IterativeSolvers

shivin9:
@jiahao 
LinearMap approach is working for lsqr as follows. It is using the function to calculate L*u instead of calculating it as a matrix multiplication. I have made some changes to LinearMaps.jl also so its current version won't work.
```
function laplace_times(next_u, u)
    for i in 2:length(u)-1
        next_u[i] = -u[i-1]+2u[i]-u[i+1]
    end
end

L = LinearMap(laplace_times, 4, ismutating=true)
b = rand(4)
lsqr(L, b)
4-element Array{Float64,1}:
 -0.28294 
  0.278895
  0.291031
 -0.286985
```

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -1,6 +1,8 @@
 import  Base: eltype, eps, length, ndims, real, size, *, \,
         A_mul_B!, Ac_mul_B, Ac_mul_B!
 
+import LinearMaps.FunctionMap
+
 export  A_mul_B
 
 # Improve readability of iterative methods
@@ -10,17 +12,22 @@ export  A_mul_B
 #### Type-handling
 """
     Adivtype(A, b)
+If A is a function map then:
+`typeof(one(eltype(b)))`
 Determine type of the division of an element of `b` against an element of `A`:
 `typeof(one(eltype(b))/one(eltype(A)))`
 """
-Adivtype(A, b) = typeof(one(eltype(b))/one(eltype(A)))
+Adivtype(A, b) = isa(A, FunctionMap) ? typeof(one(eltype(b))) : typeof(one(eltype(b))/one(eltype(A)))
 
 """
     Amultype(A, x)
+If A is a function map then:
+`typeof(one(eltype(b)))`
 Determine type of the multiplication of an element of `b` with an element of `A`:
 `typeof(one(eltype(A))*one(eltype(x)))`
 """
-Amultype(A, x) = typeof(one(eltype(A))*one(eltype(x)))
+Amultype(A, x) = isa(A, FunctionMap) ? typeof(one(eltype(b))) :typeof(one(eltype(A))*one(eltype(x)))
+
 if VERSION < v"0.4.0-dev+6068"
     real{T<:Real}(::Type{Complex{T}}) = T
     real{T<:Real}(::Type{T}) = T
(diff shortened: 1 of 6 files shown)


src/common.jl
```diff
@@ -10,17 +12,22 @@ export  A_mul_B
 #### Type-handling
 """
     Adivtype(A, b)
+If A is a function map then:
+`typeof(one(eltype(b)))`
 Determine type of the division of an element of `b` against an element of `A`:
 `typeof(one(eltype(b))/one(eltype(A)))`
 """
-Adivtype(A, b) = typeof(one(eltype(b))/one(eltype(A)))
+Adivtype(A, b) = isa(A, FunctionMap) ? typeof(one(eltype(b))) : typeof(one(eltype(b))/one(eltype(A)))
```
andreasnoack: Doesn't `FunctionMap`s have `eltype`?

andreasnoack:
I don't think it should be necessary to add LinearMaps as a dependency here. The iterative solvers should be written such that you can pass in matrix like things without knowing more about the *thing*. Personally, I'd also prefer to define a `Laplacian` type and overload `A_mul_B!` for `Laplacian` instead of defining `laplace_times `.

src/common.jl
shivin9: No, it's not there.

shivin9: No, it's not there.

shivin9:
@ChrisRackauckas and I had an in-depth discussion regarding this issue and Chris suggested that it's better to add it as a dependency. And not adding it as a dependency would lead to code redundancy.(I had tried to do it that way before).
I don't fully understand what you mean by `Laplacian` type? Do you want IterativeSolvers.jl to support `Laplacian` and other such functions intrinsically?

ChrisRackauckas:
>I don't think it should be necessary to add LinearMaps as a dependency here. The iterative solvers should be written such that you can pass in matrix like things without knowing more about the thing.

I was suggesting that all of the internals should be written for LinearMaps, and that there should just be standard dispatches so that way if what the user passes in isn't a linear map (a matrix or a function), it just gets appropriately wrapped as a LinearMap and proceeds. This would generalize the API but not inconvenience users who don't need the generality of linear maps (and should be no overhead over using a matrix or function directly). If that's how it's used, I don't see why the dependency would have to be avoided.

src/common.jl
ChrisRackauckas: Probably should be added via a small PR to FunctionMaps.jl then.

ChrisRackauckas: Actually, isn't it here? 

https://github.com/Jutho/LinearMaps.jl/blob/master/src/LinearMaps.jl#L8

eltype should work then since `FunctionMap{T,F1,F2}<:AbstractLinearMap{T}`

shivin9: Yes I found it, it's there. Will correct it in later PRs if work proceeds.

lopezm94:
Done https://github.com/Jutho/LinearMaps.jl/issues/11

shivin9:
LinearMaps has still not integrated the [fallback option](https://github.com/Jutho/LinearMaps.jl/pull/8) PR, so we may not be able to apply the complements of Linear Maps.

ChrisRackauckas:
Failing when it's not defined is fine behavior. We should try to see if there's a way to get the a lazy version of the transpose if that's possible, but "full"ing should probably be a choice since that can easily crash Julia for large sparse problems.

PR #108, opened 2017-02-22
Revert the methods added in #84

tkelman:
This is "type piracy" - extending a Base method on a signature of all Base types. Base's method tables are shared global mutable state, doing things like this can change the behavior of completely unrelated code when this package is imported.

nightly travis failure is not new, same happened at https://travis-ci.org/JuliaMath/IterativeSolvers.jl/jobs/204028681

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5-
+julia 0.5
 
 UnicodePlots
 RecipesBase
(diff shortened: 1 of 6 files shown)


andreasnoack:
cc @lopezm94

ChrisRackauckas:
This may break some things, but what this is really pointing to is that the real strategy for allowing functions is to use LinearMaps like https://github.com/JuliaMath/IterativeSolvers.jl/pull/104

PR #113, opened 2017-05-10
remove piracy on eps

tkelman:
shouldn't extend base functions on base types, as that can change the behavior of unrelated code

see if anything breaks from removing this

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,4 +4,4 @@ UnicodePlots
 RecipesBase
 SugarBLAS
 LinearMaps
-Compat 0.17.0
+Compat 0.18.0
--- a/src/common.jl
+++ b/src/common.jl
@@ -1,4 +1,4 @@
-import  Base: eltype, eps, length, ndims, real, size, *,
+import  Base: eltype, length, ndims, real, size, *,
         A_mul_B!, Ac_mul_B, Ac_mul_B!
 
 using   LinearMaps
@@ -19,12 +19,6 @@ Determine type of the multiplication of an element of `b` with an element of `A`
 `typeof(one(eltype(A))*one(eltype(x)))`
 """
 Amultype(A, x) = typeof(one(eltype(A))*one(eltype(x)))
-
-if VERSION < v"0.4.0-dev+6068"
-    real{T<:Real}(::Type{Complex{T}}) = T
-    real{T<:Real}(::Type{T}) = T
-end
-eps{T<:Real}(::Type{Complex{T}}) = eps(T)
 
 """
     randx(A, b)
--- a/src/lsqr.jl
+++ b/src/lsqr.jl
@@ -34,8 +34,8 @@ end
 #    - Eliminate printing
 #----------------------------------------------------------------------
 function lsqr_method!(log::ConvergenceHistory, x, A, b;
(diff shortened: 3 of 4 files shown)


tkelman:
looks like travis still passes on 0.5 at least

ChrisRackauckas:
It should be okay. It looks like every usage of `eps` is like:

```julia
sqrt(eps(typeof(real(b[1]))))
```

and so it actually doesn't use the `eps` overload.

tkelman:
@lopezm94 @shivin9 can you think of any reason not to make this change?

lopezm94:
In some methods it could cause errors. Would you propose doing an `eps` copy @tkelman ?

```julia
julia> using IterativeSolvers

julia> A = rand(Complex{Float64},10,10);

julia> b = rand(Complex{Float64},10);

julia> lsqr(A,b)
ERROR: MethodError: no method matching eps(::Type{Complex{Float64}})
```

Also would it be better to remove the version off of `Compat`? I can't imagine it breaking anything.

ChrisRackauckas:
Why can you `eps(typeof(real(b)))` in there as well?

tkelman:
> Also would it be better to remove the version off of Compat? I can't imagine it breaking anything.

Absolutely not. You're relying on the parametric type alias syntax for `@compat const UnrestartedHistory{T} = ConvergenceHistory{T, Void}` which was not implemented in Compat at all prior to 0.18. If anyone has Compat pinned to an earlier version than 0.18 on Julia 0.5, that would give `ERROR: syntax: expected identifier after "const"`.

lopezm94:
> Why can you eps(typeof(real(b))) in there as well?

I don't really know if not putting `real` has a purpose or someone just forgot to add it.

ChrisRackauckas:
>I don't really know if not putting real has a purpose or someone just forgot to add it.

It has a purpose. The reason is because `eps` doesn't exist for complex numbers. What you instead want is `eps` for the underlying number type of the complex number. If `b` is complex, both `eps(typeof(real(b)))` or `eps(real(typeof(b)))` will work.

ChrisRackauckas:
(Actually this reminds me that we really should have a PR for complex `eps` and `rand` in Base. Simple to implement, but I run across this frequently for very similar reasons)

tkelman:
May expect some discussion about whether `eps` should be elementwise or a magnitude. What is "next larger representable floating-point value of DataType" Complex if "larger" isn't defined? But if that were to go through and be added to Base, then it could be backported to Compat and depended on here via Compat - things that go into future versions of Julia get a type piracy pass in that way. For now I'm not sure it'd be non-controversial, so better to avoid needing it, or make a local version instead of globally extending the Base function.

lopezm94:
> The reason is because eps doesn't exist for complex numbers
Yeah this part I understood of because the error I displayed. I meant something like using `real` would transform sometimes the inner type of `Complex` instead of extracting it, but I just tested some types now and it appears to extract it correctly. I'll add `real` to all the calls and merge.

tkelman:
sorry didn't notice your edit, and wasn't sure exactly where the calls that needed it were

since tests didn't catch that usage it may be good to add a complex-valued regression test, but other than your snippet above I'm not sure what that should look like

Issue #109, opened 2017-04-05
Incorrect `gmres`?


panlanfeng:
In the following example
```julia
import IterativeSolvers
n=30;
A = randn(n,n); A= A.+A'; 
x = rand(n);
b = A*x; 
A*IterativeSolvers.gmres(A, b) .- b
```
It only work for `n` less than 20. This is the case for both v0.2.2 and master.

Am I using it the wrong way?

panlanfeng:
I see. It is correct if I set `restart=length(b)+1`.
So the problem should be the default value of `restart` is too small.
For reference, the R package `pracma` provides the same function and its default `restart` is `length(b)+1`.

haampie:
You're completely right, there is a bug in the current GMRES implementation where the best approximation is not incremented at restart, but overwritten. Let me write a one character PR for this ;)

haampie:
This can be closed. Fixed by https://github.com/JuliaMath/IterativeSolvers.jl/pull/119/commits/cd2ad63d91148f4ae679ecd4fcafb6f01d796908

PR #119, opened 2017-05-23
Update, not reset the solution at restart

haampie:
The current restarted GMRES implementation has a bug #109 where the approximate solution is overwritten rather than incremented at restart.

Below is the convergence history for a simple 1D Poisson matrix. Since GMRES selects the minimal residual from the Krylov search space, the normed residual should not increase like this.

not_optimal

Here the residual for the same problem after the fix:

optimal

To reproduce the above:

```julia
poisson1d(n) = spdiagm( (ones(n - 1), -2 * ones(n), ones(n - 1)), (-1, 0, 1) )
A = poisson1d(200)
x = rand(200)
b = A * x
x, history = gmres(A, b; log = true, maxiter = 100, restart = 20)
plot(history.data[:resnorm], yscale = :log10)
```

### DIFF
--- a/src/gmres.jl
+++ b/src/gmres.jl
@@ -114,7 +114,7 @@ function gmres_method!(log::ConvergenceHistory, x, A, b;
 
         @eval a = $(VERSION < v"0.4-" ? Triangular(H[1:N, 1:N], :U) \ s[1:N] : UpperTriangular(H[1:N, 1:N]) \ s[1:N])
         w = a[1:N] * K
-        @blas! x = solve(Pr,w) #Right preconditioner
+        @blas! x += solve(Pr,w) #Right preconditioner
 
         if (rho<tol) | ((macroiter-1)*restart+N >= maxiter)
             setconv(log, rho<tol)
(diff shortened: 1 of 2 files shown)


src/gmres.jl
```diff
@@ -114,7 +114,7 @@ function gmres_method!(log::ConvergenceHistory, x, A, b;
 
         @eval a = $(VERSION < v"0.4-" ? Triangular(H[1:N, 1:N], :U) \ s[1:N] : UpperTriangular(H[1:N, 1:N]) \ s[1:N])
         w = a[1:N] * K
-        @blas! x = solve(Pr,w) #Right preconditioner
+        @blas! x += solve(Pr,w) #Right preconditioner
```
lopezm94: Thanks for spotting that! I totally missed it when translating. Could you also add a test case to [this](https://github.com/JuliaMath/IterativeSolvers.jl/blob/master/test/runtests.jl#L59) file? Just checking if it converges correctly.

haampie: Will do!

haampie:
I have added a test for checking whether the residual norm is non-increasing

PR #124, opened 2017-06-07
Optimizing cg.jl in terms of reduced memory allocations

shivin9:
cg.jl had several memory leaks arising from https://github.com/JuliaMath/IterativeSolvers.jl/blob/master/src/cg.jl#L45
and 
https://github.com/JuliaMath/IterativeSolvers.jl/blob/master/src/cg.jl#L54.
``` julia
julia> @time x = cg(A, rhs; maxiter=2000);
  0.073056 seconds (12.04 k allocations: 31.635 MB, 14.14% gc time)
```
In this PR I have defined mutating functions for these functionalities and now we have
``` julia
julia> @time cg(A, rhs; maxiter=2000);
  0.051681 seconds (8.04 k allocations: 714.328 KB)
```

Please note that I have found many such areas of optimizations in gmres.jl, lancoz.jl etc. but due to want of time I won't be able to address those issues. 
Maybe @haampie can fix these during the summer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # mkdocs site
 /site
 /docs/build
+*.mem
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,3 +5,4 @@ RecipesBase
 SugarBLAS
 LinearMaps
 Compat 0.18.0
+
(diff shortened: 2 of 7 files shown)


shivin9:
sorry about this, will issue a correct PR with updated master soon

PR #128, opened 2017-06-11
cg_mem_fix (missed this commit before)

shivin9:
Using nextvec! to decrease further consumption of memory. It was there in my original fix but I missed this while cherry picking it into master :P.


lopezm94:
Thanks! Try to not make PRs with your master branch, otherwise it is a bit troublesome to edit because it will always conflict with the local master branch whenever your PR is pulled.

PR #130, opened 2017-06-15
Reduce memory usage in CG

haampie:
- Pre-allocates the vectors, improvement shown below.
- Removes the dependency of CG on KrylovSubspace (makes little sense for a 3-term recurrence method IMHO). Also: the KrylovSubspace methods are the main reason for inefficient memory usage. For the moment it was easier to fix just CG.
- Fixes a bug where the CG method starts while the initial guess is the actual solution (I still have to add a test for this)
- Adds a simple benchmark for the CG method
- Relies always on `A_mul_B!` for matrix-vector products. Since we're switching to LinearMaps this shouldn't be a problem.
- Also: removes the check whether all elements of `b` are zero. I would say CG should not be responsible for this.

Benchmark results:

```
LinearSystemsBench.cg()
```

*This PR*
```
BenchmarkTools.Trial:
  memory estimate:  45.80 MiB
  allocs estimate:  437
  --------------
  minimum time:     2.561 s (0.16% GC)
  median time:      2.580 s (0.27% GC)
  mean time:        2.580 s (0.27% GC)
  maximum time:     2.598 s (0.38% GC)
  --------------
  samples:          2
  evals/sample:     1
```

*Current implementation*
```
  BenchmarkTools.Trial:
  memory estimate:  1.53 GiB
  allocs estimate:  1041
  --------------
  minimum time:     3.173 s (9.28% GC)
  median time:      3.246 s (11.98% GC)
  mean time:        3.246 s (11.98% GC)
  maximum time:     3.320 s (14.57% GC)
  --------------
  samples:          2
  evals/sample:     1
```

### DIFF
--- a/benchmark/benchmark-linear-systems.jl
+++ b/benchmark/benchmark-linear-systems.jl
@@ -1,0 +1,44 @@
+module LinearSystemsBench
+
+import Base.A_ldiv_B!, Base.\
+
+using BenchmarkTools
+using IterativeSolvers
+
+# A DiagonalMatrix that doesn't check whether it is singular in the \ op.
+immutable DiagonalPreconditioner{T}
+    diag::Vector{T}
+end
+
+function A_ldiv_B!{T}(y::AbstractVector{T}, A::DiagonalPreconditioner{T}, b::AbstractVector{T})
+    for i = 1 : length(b)
+        @inbounds y[i] = A.diag[i] \ b[i]
+    end
+    y
+end
+
+(\)(D::DiagonalPreconditioner, b::AbstractVector) = D.diag .\ b
+
+function posdef(n)
+    A = SymTridiagonal(fill(2.01, n), fill(-1.0, n))
+    b = A * ones(n)
+    return A, b
+end
+
+function cg(; n = 1_000_000, tol = 1e-6, maxiter::Int = 200)
+    A, b = posdef(n)
+    P = DiagonalPreconditioner(collect(linspace(1.0, 2.0, n)))
+
+    println("Symmetric positive definite matrix of size ", n)
+    println("Eigenvalues in interval [0.01, 4.01]")
+    println("Tolerance = ", tol, "; max #iterations = ", maxiter)
+    
+    # Dry run
+    initial = rand(n)
+    IterativeSolvers.cg!(copy(initial), A, b, Pl = P, maxiter = maxiter, tol = tol, log = false)
+
+    # Actual benchmark
+    @benchmark IterativeSolvers.cg!(x0, $A, $b, Pl = $P, maxiter = $maxiter, tol = $tol, log = false) setup=(x0 = copy($initial))
+end
+
+end
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -1,66 +1,80 @@
 export cg, cg!
 
-####################
-# API method calls #
-####################
-
-cg(A, b; kwargs...) = cg!(zerox(A,b), A, b; kwargs...)
+cg(A, b; kwargs...) = cg!(zerox(A, b), A, b; kwargs...)
 
 function cg!(x, A, b;
-             tol::Real=size(A,2)*eps(), maxiter::Integer=size(A,2),
-             plot=false, log::Bool=false, kwargs...
-             )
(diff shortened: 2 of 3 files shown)


benchmark/benchmark-linear-systems.jl
```diff
@@ -0,0 +1,44 @@
+module LinearSystemsBench
+
+import Base.A_ldiv_B!, Base.\
+
+using BenchmarkTools
+using IterativeSolvers
+
+# A DiagonalMatrix that doesn't check whether it is singular in the \ op.
```
KristofferC: What is the performance difference of this compared to the standard `Diagonal`?

haampie:
Hmm, I missed the follow up of @shivin9 in #128, but still I think it is better to not use the KrylovSubspace functions and shoot just for `A_mul_B!`. CG does Krylov subspace things implicitly, and it feels hacky to initialize a KrylovSubspace of order 1.

timholy:
> Also: removes the check whether all elements of b are zero. I would say CG should not be responsible for this.

Since there's a test for this, it's fine to change as long as it passes the test. Without that check it used to return all-NaN, but presumably your fix of the starting-point-at-solution bug fixed this.

haampie:
I think the NaN result is a bug in the current implementation because the initial residual is not checked. With `b = zeros(n)` and `x0 = zeros(n)` you start out with a solution, so the iterations should not even start. The current impl. only inspect the residual after the first iteration, this PR does it before the first iteration.

If `b = zeros(n)` while `x0 = rand(n)`, then CG works just fine.

lopezm94:
Removing `KrylovSubspace` should probably be in a new issue and left intact in this PR. Some solvers in this package already accept `KrylovSubspace` as input, the intention with making the `cg` implementation dependant was to enable it as input in the future. So if we are going to remove `KrylovSubspace`, it should be generalized in the whole package.

If I remember correctly there was an idea to remove `KrylovSubspace` from the package, right @jiahao? Something along the lines of using it only as a store of vectors and providing normalization functions. I could open a new issue about this. This could also improve readability for people who are more accustomed to see the matrix operations inside the solvers.

lopezm94:
@haampie can you leave KrylovSubspace dependency as it was and rebase? To not delay this PR's merge unnecessarily and then we can discuss KrylovSubspace in general for the package.

haampie:
This PR doesn't remove the `KrylovSubspace` type, it removes its usage in `cg.jl`, so this PR can be merged I guess.

lopezm94:
Yeah the issue is about removing it's usage in cg, not about removing the type from the package. As I explained before:

'''
Some solvers in this package already accept KrylovSubspace as input, the intention with making the cg implementation dependant was to enable it as input in the future. So if we are going to remove KrylovSubspace, it should be generalized in the whole package.
'''

So the intention was to make cg accept krylovsubspace as input in the future. If we remove it then it can't be done and it should be in a separate issue.

haampie:
Alright, fixed that

barche:
Just for reference, I did a comparison with Trilinos.jl, resulting in the following for CG:
* Your code: 2.7 s
* Trilinos: 1.633 s

I attempted to reproduce your benchmark as accurately as possible here:
https://github.com/barche/Trilinos.jl/blob/master/benchmarks/belos.jl

ChrisRackauckas:
Thanks @barche for taking the time to do that. It'll serve as a good performance goal to keep in mind! That shows we're already pretty close to Trillinos which is awesome!

KristofferC:
Is Trilinos multithreading the sparse matrix vector multiplication? If it is, one could try using something like https://github.com/JuliaSparse/MKLSparse.jl or https://github.com/dcjones/RecursiveSparseBlocks.jl to see if it is the matvec that makes the difference.

haampie:
@barche did you run Julia with multithreading enabled? In my experience the Julia version runs somewhat faster if you do `BLAS.set_num_threads(1)` because of all the BLAS level 1 operations.

barche:
@KristofferC No, Trilinos was not multithreading. I'll do some more tests using that when I get back to one of my desktop machines.

@haampie I kept the Julia settings at their default, but switching off threading made no noticeable difference on this 2012 Macbook Air.

PR #131, opened 2017-06-28
Fix GMRES memory usage

### LINKED ISSUE #127: Guaranteed memory usage (in GMRES)
The current implementation of GMRES allocates too much memory. We should be able to guarantee that memory usage is fixed as `O(restart * n)`, where `n` is the height of the matrix. 

As a reference, I've coded up [an improved GMRES version](https://gist.github.com/haampie/895056490d3eb22c30e37ca6bf347abb#file-gmres-jl-L8-L45). The main difference is that it [pre-allocates](https://gist.github.com/haampie/895056490d3eb22c30e37ca6bf347abb#file-gmres-jl-L55) the `V` matrix and reuses that without throwing vectors away. Its implementation is super basic, it does not have preconditioners and does not factorize the Hessenberg matrix. 

Then I [benchmarked](https://gist.github.com/haampie/895056490d3eb22c30e37ca6bf347abb#file-gmres-jl-L129-L145) against the version of IterativeSolvers.jl. GMRES does not converge for the specific matrix, so that you can compare results for a large number of iterations.

For `n = 5000` and `1500` iterations with restarts after `15` iterations, this is what I get:

**IterativeSolvers.jl**

```
BenchmarkTools.Trial:
  memory estimate:  423.87 MiB
  allocs estimate:  98640
  --------------
  minimum time:     579.899 ms (3.20% GC)
  median time:      581.539 ms (3.19% GC)
  mean time:        581.258 ms (3.24% GC)
  maximum time:     582.504 ms (3.41% GC)
  --------------
  samples:          9
  evals/sample:     1
```

**Improved version**

```
BenchmarkTools.Trial:
  memory estimate:  13.54 MiB
  allocs estimate:  20492
  --------------
  minimum time:     479.069 ms (0.00% GC)
  median time:      481.632 ms (0.31% GC)
  mean time:        481.409 ms (0.17% GC)
  maximum time:     482.786 ms (0.31% GC)
  --------------
  samples:          11
  evals/sample:     1
```

So, memory usage is reduced a lot, just like the number of allocations. Maybe it can be improved upon.

Surprisingly it is faster as well... Maybe factorizing the Hessenberg matrix on the go is slower than [a simple \\](https://gist.github.com/haampie/895056490d3eb22c30e37ca6bf347abb#file-gmres-jl-L103).

With a smaller matrix I got some more samples (~10000):

```
> current, improved = benchmark(n = 500)
> judge(median(improved), median(current))
BenchmarkTools.TrialJudgement:
  time:   -41.32% => improvement (5.00% tolerance)
  memory: -73.43% => improvement (1.00% tolerance)
```

TL;DR:
- Pre-allocate the `V` matrix and use indices to reduce the number of allocations
- Maybe don't even do the Hessenberg QR-decomposition during the iterations

haampie:
This PR contains a replacement for the current GMRES implementation aimed to fix #127. Feedback is welcome :)

What's different w.r.t. the current implementation:

1. Pre-allocates (almost) all vectors. The only problem is how to deal with preconditioners efficiently
2. Computes the residual without computing the QR decomposition of the Hessenberg matrix. This way you don't have to store the Given's rotations + the inner loop of GMRES gets cleaner
3. Adds a special Hessenberg type (different from LinAlg.Hessenberg) that takes care of the small least-squares problem via Given's rotations + UpperTriangular.
3. ~Uses UnsafeVectorView for non-allocating, small views of vectors.~ Bug prone, so removed.
4. Only append to the ConvergenceHistory when logging is turned on
5. Adds an Identity type as preconditioner so that without preconditioning we can do a non-allocating expansion of the Krylov subspace (i.e. `V[k + 1] = A * V[k]` will not allocate`)
6. Adds some benchmarks
7. Changes `V` to a `Matrix` rather than `Vector{Vector}` so that BLAS-2 can be exploited in the update of the solution `x += Vy`.

Todo:
- [ ] Rename to `gmres`
- [x] Add different orthogonalization strategies~

Running `LinearSystemsBench.gmres()` gives something like:

```
julia> improved
BenchmarkTools.Trial:
  memory estimate:  13.75 MiB
  allocs estimate:  313
  --------------
  minimum time:     11.454 s (0.00% GC)

julia> old
BenchmarkTools.Trial:
  memory estimate:  1.26 GiB
  allocs estimate:  13378
  --------------
  minimum time:     13.568 s (8.59% GC)
```


src/hessenberg.jl
```diff
@@ -0,0 +1,39 @@
+import Base.LinAlg: Givens, givensAlgorithm
+
+mutable struct Hessenberg{T<:AbstractMatrix}
+    H::T # H is assumed to be Hessenberg of size (m + 1) × m
+end
+
+@inline Base.size(H::Hessenberg, args...) = size(H.H, args...)
+
+function solve!(H::Hessenberg, rhs)
```
andreasnoack: We usually call these `A_ldiv_B!`

src/hessenberg.jl
```diff
@@ -0,0 +1,39 @@
...
+
+        # Right hand side
+        tmp = -s * rhs[i] + c * rhs[i + 1]
+        rhs[i] = c * rhs[i] + s * rhs[i + 1]
+        rhs[i + 1] = tmp
+    end
+
+    # Solve the upper triangular problem.
+    U = UpperTriangular(view(H.H, 1 : width, 1 : width))
+    U \ UnsafeVectorView(rhs, 1 : width)
```
andreasnoack: Why not in-place solve here?

src/hessenberg.jl
```diff
@@ -0,0 +1,39 @@
+import Base.LinAlg: Givens, givensAlgorithm
+
+mutable struct Hessenberg{T<:AbstractMatrix}
```
andreasnoack: We should consider upstreaming this to Base

src/improved_gmres.jl
```diff
@@ -0,0 +1,211 @@
...
+    @blas! arnoldi.V[1] *= one(T) / β
+    β
+end
+
+@inline function init_residual!(r::Residual{numT, resT}, β) where {numT,resT}
+    r.accumulator = one(resT)
+    r.β = β
+end
+
+function solve_least_squares!(arnoldi::ArnoldiDecomp{T}, β, k::Int) where {T}
```
andreasnoack: Maybe a more precise name here since it is not solving arbitrary least squares problems

src/hessenberg.jl
haampie: I wasn't completely sure, because this method mutates the `H` matrix as well

src/hessenberg.jl
haampie: I'm doing that now in the latest commit, but it's a bit awkward because it's a least-squares problem. The Hessenberg matrix is not square, so it comes down to having the solution in the first `k` elements of `rhs` and having a last element at index `k + 1` not being part of the solution.

src/hessenberg.jl
andreasnoack: I think we should stick with `A_ldiv_B!` here even though it invalidates the factorization. Probably a good idea to add a comment, though.

andreasnoack:
I think we a ready for this one so if you finish it up we can go ahead and merge

PR #133, opened 2017-07-02
Fix CG without preconditioning

haampie:
This PR makes CG without preconditioning ~30% more efficient in wall-clock time. Optimizations in the non-preconditioned case are:

1. The residual `norm(r)` is implicitly available, so it doesn't have to be computed.
2. `dot(r, c)` becomes `dot(r, r)`; since this operation is memory-bound, it's twice as fast.
3. No need to copy the `r` vector to `c` when "preconditioning".

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -7,60 +7,76 @@ function cg!(x, A, b;
     maxiter::Integer = min(20, size(A, 1)),
     plot = false,
     log::Bool = false,
+    Pl = Identity(),
     kwargs...
 )
     (plot & !log) && error("Can't plot when log keyword is false")
-    K = KrylovSubspace(A, length(b), 1, Vector{Adivtype(A,b)}[])
-    init!(K, x)
     history = ConvergenceHistory(partial = !log)
     history[:tol] = tol
-    reserve!(history, :resnorm, maxiter)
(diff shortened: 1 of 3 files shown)


src/cg.jl
```diff
@@ -7,80 +7,161 @@ function cg!(x, A, b;
...
         ρ_prev = ρ
         ρ = dot(c, r)
-        β = -ρ / ρ_prev
+        β = ρ / ρ_prev
 
-        # u := r - βu (almost an axpy)
-        @blas! u *= -β
-        @blas! u += one(T) * c
+        # u := c + βu (almost an axpy)
+        @inbounds @simd for i = 1 : length(u)
```
andreasnoack: I think we should avoid explicit loops on the vectors. This will make it impossible to use the solver with a distributed array

src/cg.jl
```diff
@@ -7,80 +7,161 @@ function cg!(x, A, b;
...
+
         verbose && @printf("%3d\t%1.2e\n", iter, last_residual)
+
+        # Stopping condition
+        if last_residual ≤ reltol || iter > maxiter
+            break
+        end
+
+        # u := r + βu (almost an axpy)
+        @inbounds @simd for i = 1 : length(u)
```
andreasnoack: Same as above

haampie:
Ok, fixed the simd part and finally all tests are green

PR #134, opened 2017-07-06
BiCGStab(ℓ)

haampie:
This is still a work in progress version of BiCGStab(ℓ). It is based on the original article, uses BLAS and includes left preconditioning.

Todo:

- [x] Incorporate the convergence history stuff.
- [x] ~Improvement of [1]~
- [x] ~Improvement of [2]~

[1] Sleijpen, Gerard LG, and Henk A. Vorst. "Maintaining convergence properties of BiCGstab methods in finite precision arithmetic." Numerical Algorithms 10.2 (1995): 203-223.
[2] Sleijpen, G. LG, and Henk A. van der Vorst. "Reliable updated residuals in hybrid Bi-CG methods." Computing 56.2 (1996): 141-163.

### DIFF
--- a/benchmark/benchmark-linear-systems.jl
+++ b/benchmark/benchmark-linear-systems.jl
@@ -5,6 +5,8 @@ import Base.A_ldiv_B!, Base.\
 using BenchmarkTools
 using IterativeSolvers
 
+include("../test/advection_diffusion.jl")
+
 # A DiagonalMatrix that doesn't check whether it is singular in the \ op.
 immutable DiagonalPreconditioner{T}
     diag::Vector{T}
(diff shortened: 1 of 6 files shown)


haampie:
At the moment I don't think it is completely necessary to implement the improvements from the papers. The thing is there is no standard implementation of BiCGStab(ℓ) and there's a lot of freedom.

The current implementation works fine for moderate values of `ℓ`. if someone wants to use large `ℓ` and still have enough precision, then it's a good idea to switch to a fancy implementation that solves the polynomial part using Gram-Schmidt rather than normal equations and once in a while recomputes the exact residual.

Issue #129, opened 2017-06-15
Visitor pattern for convergence history


haampie:
Currently there are two or three options like `plot`, `log` and `verbose` for the solvers. In fact these are all different ways to show progress of the algorithm. I think it is a good design choice to separate output about progress completely from the algorithms themselves.

One way to go is to use the visitor pattern: let the user pass in a callback, and each iteration call it with the current convergence history results. Then you could do stuff like this:

```julia
p = plot(yscale = :log10)
plot_on_the_go = t::ConvergenceHistory -> push!(p, t.iterations, last(t[:resnorm]))
cg(A, b, visitor = plot_on_the_go)
```

Is there a drawback to this?

haampie:
The problem with the above is that the algorithm can't know what data the user needs. Maybe a `ConvergenceHistory` object isn't enough:

In GMRES you might want to access the Hessenberg matrix to compute Ritz values at the end of a "macro" iteration; you could use approximate eigenvectors to apply deflation techniques in a second problem you need to solve. Or you might want to implement a multishift version of GMRES where you solve (A - tI)x = b for multiple `t`'s, reusing the search subspace from the `Ax=b` problem. In BiCGStab(l) you might want to compare how the computed residual actually differs from the real residual (so you need access to intermediate solutions `x`). Et cetera.

All this should be possible without changing the algorithm itself or copy / pasting it into a new one.

Now @jiahao started with the iterator idea for CG, and I currently think that is the best way to make it possible. So basically you have some low-level helper function to construct an iterable:

```
iterable = cg_iterable(A, b, ...)
```

And then let the algorithm run as

```
for (iteration, residual) = enumerate(iterable)
    println("Iteration ", iteration, " |r| = ", residual)
end
```

The `iterable` gives the user access to `x` and `r` et cetera:

```
for res = iterable
  surface(reshape(iterable.x, 128, 128))
end
```

The original API can then just be a thin wrapper around the iterator:

```
function cg!(x, A, b; kwargs...)
  iterable = cg_iterable(x, A, b; kwargs...)

  for item = iterable

  iterable.x
end
```

ChrisRackauckas:
[...]

PR #137, opened 2017-07-11
CG with iterators

haampie:
First example implementing #129.

The good thing is that logging & printing is completely separated from the algorithm, which means less noise.

Benchmark show no regressions


haampie:
Unrelated question @lopezm94: can we modify the coveralls thing a bit? It produces huge reports that say basically nothing, and then fail because there is a 0.03% reduction in coverage.

andreasnoack:
Needs a rebase now

PR #139, opened 2017-07-11
Power method iterators

haampie:
As part of getting rid of `KrylovSubspace` and adding iterators as suggested in #129, here the fix for the power method.

There is a breaking change: the shift-and-invert matrix is not constructed in the method itself (and for good reason, otherwise you could just use RQI and get better convergence at the same computational costs per step). It is assumed the user provides a linear map `B` supporting `A_mul_B!`, which has the effect `B * x = inv(A - σI) * x` where σ is the shift. The only convenience `invpowm` gives over `powm` is that the eigenvalue is transformed for you.

For example:

```
using LinearMaps
σ = 1.0 + 1.3im
A = rand(Complex128, 50, 50)
F = lufact(A - UniformScaling(σ))
Fmap = LinearMap((y, x) -> A_ldiv_B!(y, F, x), 50, Complex128, ismutating = true)
λ, x = invpowm(Fmap, shift = σ, tol = 1e-4, maxiter = 200)
```

You might argue `A_ldiv_B!` should be used in `invpowm`, but the thing is someone could just as well have an explicit inverse matrix `inv(A -  σI)`, and then `A_mul_B!` should be used.

Lastly, the method now stores exactly 3 vectors, and those are pre-allocated; the current behaviour is allocating a new matrix each iteration, together with multiple new vectors.

### DIFF
--- a/src/simple.jl
+++ b/src/simple.jl
@@ -1,3 +1,5 @@
+import Base: start, next, done
+
 #Simple methods
 export powm, invpowm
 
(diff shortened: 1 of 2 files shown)


src/simple.jl
```diff
@@ -1,96 +1,115 @@
...
 # API method calls #
 ####################
 
-function invpowm(A;
-    x=nothing, shift::Number=0, tol::Real=eps(real(eltype(A)))*size(A,2)^3,
-    maxiter::Int=size(A,2), log::Bool=false, kwargs...
-    )
-    K = KrylovSubspace(A, 1)
-    x==nothing ? initrand!(K) : init!(K, x/norm(x))
+function powm(A; kwargs...)
```
andreasnoack: Not related to this PR but we might want to consider renaming this function to something that looks less like a matrix power function.

Issue #144, opened 2017-07-14
Lanczos implementation not useful in practice


haampie:
There are multiple issues with the current implementation of the Lanczos method for the Hermetian eigenvalue problem, which makes it actually unusable in practice:

1. No re-orthogonalization: due to loss of orthogonality of the Krylov basis vectors, eigenvalues are detected multiple times (and therefore some aren't found):

```
> srand(1)
> A = rand(100, 100)
> A = A + A'
> eigvals(A)[end - 10 : end]
11-element Array{Float64,1}:
   5.74519
   5.88254
   6.3581
   6.40497
   6.56969
   6.83121
   7.01442
   7.21099
   7.48439
   7.58956
 100.818
> eiglancz(A)[end - 10 : end]
11-element Array{Float64,1}:
   7.48439
   7.58956
 100.818
 100.818
 100.818
 100.818
 100.818
 100.818
 100.818
 100.818
 100.818
```
2. The `maxiter` keyword is not passed on to the actual method
3. The stopping criterion is not useful: often you want to compute a couple eigenvalues near a target, and while other Ritz values aren't converged, the ones near the target may have. The stopping target should incorporate that.
4. The current implementation cannot deal with multiple eigenvalues: only one vector in the null space of (A - sI) is found, while the dimension can be larger than 1.

andreasnoack:
Fixed by #146

PR #146, opened 2017-07-14
Remove Lanczos

haampie:
See the discussion in #144. It seems to take quite a bit of work to make the Lanczos method useful in practice (reorthogonalization, storing the basis vectors, restarts, block-stuff for multiple eigenpairs), and since it isn't at the moment, it might be better for the credibility of this package to remove it.

Of course git history remains if someone wants to pick this method up.


ChrisRackauckas:
Wouldn't it be easier to just leave it and document it as a work in progress?

andreasnoack:
I think it would be as easy, if not easier, to start from scratch when somebody wants to write a new implementation. I agree with @haampie that it would be better for the credibility of the package not to have this implementation since it would give wrong results for most problems.

haampie:
Btw, @jiahao already gone through most of the subtleties of a proper Lanczos implementation in svdl.jl. So if someone picks up this again, that'll be a fine reference as well.

PR #148, opened 2017-07-21
MINRES

haampie:
MINRES is basically full GMRES exploiting symmetry of the matrix so that it does short recurrences and stores only 6 vectors, requiring constant work per iteration. The current implementation requires 1 MV-product, 5 axpy's, 1 dot, 1 copy, and 3 scale!'s. 

Maybe 2 axpy's can be translated into 1 MV-product, and a `copy` and `scale!` can maybe be combined, but at any rate the current implementation is already nearly optimal.

### DIFF
--- a/benchmark/benchmark-linear-systems.jl
+++ b/benchmark/benchmark-linear-systems.jl
@@ -77,4 +77,12 @@ function bicgstabl()
     b1, b2, b3, b4
 end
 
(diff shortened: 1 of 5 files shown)


haampie:
Another option could be to support skew-Hermetian matrices `A = -A'` as well, that requires just one additional minus sign.

Issue #117, opened 2017-05-17
Code broken on 0.3.0 release


una-dinosauria:
I have a small project that uses LSQR from this package at https://github.com/una-dinosauria/local-search-quantization/. Under release 0.3.3, one of my demos broke unexpectedly.

I have a small example that shows the code breaking.

```julia
using IterativeSolvers, JLD

fname = "IterativeSolvers.jld"
C = load(fname, "A")
b = load(fname, "b")

x, _ = IterativeSolvers.lsqr( C, b )
@show x
```

With https://github.com/JuliaMath/IterativeSolvers.jl/commit/0b4d77d542ef7b08ea9ea3cc57ea314acc2c6d4b the code works like a charm:
```
[...]

PR #150, opened 2017-07-25
Remove FactCheck, use Base.test and simplify tests

haampie:
This touches all tests, so a bit of a scary PR.

It removes `FactCheck` and moves to Julia's native `Base.Test`

It also simplifies a lot of tests and removes some warnings in 0.6 about broadcasting.

With regard to the `lsmr.jl` test, it's quite elaborate (the types & methods in the test itself almost need tests), but the actual thing that is being tested is whether or not lsmr support custom matrix and vector types. I think checking for custom matrix types can just as well be done by wrapping a matrix in a `LinearMap`, rather than writing hard to follow implicit matrix-vector product routines.

Concerning the custom vector types: that's something we need to test not just for `lsmr`, but for all methods. For instance using `DistributedVector`.

This PR also removes the test with the `test/IterativeSolvers.jld` which is basically a huge matrix (see #117).

Lastly, the tests in `history.jl` should probably not be smoke tests like they are now. I think it makes more sense to just unit test a `ConvergenceHistory` object.


haampie:
It turns out I accidentally broke support for Julia 0.5 in multiple ways:

- The new version of LinearMaps.jl drops 0.6 support
- `@test [expr] [kwargs]` syntax is not backwards compatible
- The GMRES test with preconditioners and sparse matrices was broken, and after a fix does not pass on Julia 0.5. It accidentally converted sparse to dense via `sprand(...) + eye(...)`, which I have fixed to `sprand(...) + speye(...)`. And only then the tests failed on Julia 0.5 because `A_ldiv_B!(::UmfpackLU, ::SubArray)` was not implemented. Current solution is to do the test conditionally on the version number.

test/gmres.jl
```diff
@@ -1,91 +1,65 @@
...
-    R = sprandn(n,n,0.5)
-    b = convert(Vector{T}, randn(n))
-    if T <: Complex
-        A += im*sprandn(n,n,0.5)
-        L += im*sprandn(n,n,0.5)
-        R += im*sprandn(n,n,0.5)
-        b += im*randn(n)
-    end
+@testset "SparseMatrixCSC{$T}" for T in (Float64, Complex128)
+    A = sprand(T, n, n, 0.5) + speye(T, n, n)
```
andreasnoack: Use `I` here instead of `speye`

haampie: Nice, I'll use that throughout the code then

test/cg.jl
```diff
@@ -1,45 +1,44 @@
...
+        # All-zeros rhs should give all-zeros lhs
+        x0 = cg(A, zeros(T, n))
+        @test x0 == zeros(T, n)
+    end
 end
 
-context("Sparse Laplacian") do
-    A = poisson_matrix(Float64, 32, 3)
+@testset "Sparse Laplacian" begin
+    A = poisson_matrix(Float64, 10, 3)
```
andreasnoack: Why change the size?

haampie: At the moment is generates a matrix of size 32^3 = 32.768, which is rather large for a test.

src/cg.jl
```diff
@@ -118,17 +118,15 @@ function cg_iterator!(x, A, b, Pl = Identity();
...
 
-    # Stopping criterion
-    ρ = one(residual)
-
     # Return the iterable
     if isa(Pl, Identity)
         return CGIterable(A, x, b,
             r, c, u,
-            reltol, residual, ρ,
+            reltol, residual, residual,
```
haampie: Whoops, yeah, that seems to be an error. I'll fix it

PR #152, opened 2017-07-26
Drop 0.5 support

haampie:
Since LinearMaps already dropped support for Julia 0.5, and only critical fixes are backported to 0.5, it might be worthwhile to drop support for 0.5 altogether. 

This makes it also possible to silence the number of deprecation warnings.


andreasnoack:
Also need to adjust the `REQUIRE` file.

PR #153, opened 2017-07-27
Modernize syntax

haampie:
Just use the Julia 0.6 syntax everywhere.

One bug fix in this PR: the stationary solvers used `SingularError` which is probably removed long ago. I'm now using `Base.LinAlg.SingularException` and added a test for it.

### DIFF
--- a/src/history.jl
+++ b/src/history.jl
@@ -51,7 +51,7 @@ ploted as series and matrices as scatterplots.
 `Base`: `getindex`, `setindex!`, `push!`
 
 """
-type ConvergenceHistory{T,K}
+mutable struct ConvergenceHistory{T,K}
     mvps::Int
     mtvps::Int
     iters::Int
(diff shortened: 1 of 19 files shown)


src/history.jl
```diff
@@ -256,8 +256,8 @@ nrests(ch::RestartedHistory) = Int(ceil(ch.iters/ch.restart))
 Determine whether a collection `x` is plotable. Only vectors and matrices are
 such objects.
 """
-plotable{T<:Real}(::VecOrMat{T})::Bool = true
```
andreasnoack: Is it on purpose that the return type annotations are removed here?

src/stationary.jl
```diff
@@ -41,7 +43,7 @@ function jacobi_method!(log::ConvergenceHistory, x, A::AbstractMatrix, b;
 			for j=[1:i-1;i+1:n]
 				xi += A[i,j]*xold[j]
 			end
-			A[i,i]==0 && throw(SingularError())
+			A[i,i]==0 && throw(SingularException(1))
```
andreasnoack: Maybe `SingularException(i)`

src/history.jl
haampie: To be honest I didn't know how to combine `where` syntax with return type annotations. But I don't think we need explicit return type annotations for such a simple function anyways.

andreasnoack: Just asked and it seems to be a bug when using the "assignment form" of function definitions. I agree that it should really matter here.

PR #156, opened 2017-07-31
Jacobi, Gauss-Seidel & SOR iteration for SparseMatrixCSC using iterators

### LINKED ISSUE #147: Stationary methods for sparse matrices
There are a couple issues with current implementations of stationary methods like `jacobi` and `gauss_seidel`:

1. Sparsity of the matrix should be exploited, so that each iteration has `O(nnz(A))` complexity rather than `O(n²)`.
2. Indexing `A[i, j]` should be avoided for sparse matrices
3. Jacobi should allocate only one additional vector, and Gauss-Seidel can do without allocations (fully in-place).

Methods like Jacobi and Gauss-Seidel split a matrix in two `A = B - C` (where B is for instance the diagonal of A in the case of Jacobi). Maybe our code should reflect that idea by wrapping the `A` matrix in some type. At the very least a bit of indirection makes it possible to treat sparse matrices differently. 

```
n = 1_000
A = speye(n) + sprand(n, n, 10 / n)
b = ...
B = Diagonal(A)
C = -OffDiagonal(A)
x = zeros(n)
tmp = similar(x)

for i = 1 : maxiter
  A_mul_B!(tmp, C, x) # out of place
  axpy!(1.0, b, tmp)
  A_ldiv_B!(x, B, tmp)
end
```

And Gauss-Seidel

```
B = Diagonal(A) - LowerTriangular(A)
C = -UpperTriangular(A)

for i = 1 : maxiter
  A_mul_B!(C, x) # in-place
  axpy!(1.0, b, x)
  A_ldiv_B!(B, x) # in-place
end
```

At this point `UpperTriangular` converts sparse matrices to dense matrices, and `triu` copies. Both are not ideal.

haampie:
Unfortunately the simple stationary methods like Gauss-Seidel and SOR are a bit awkward to implement for column-major (sparse) matrices. 

My goal with this PR is to support stationary methods without copying or transforming the underlying SparseMatrixCSC, while still exploiting in-place updates as much as possible.

Todo:

- [x] Tests
- [x] Combine this with current implementation for dense matrices
- [x] Symmetric SOR
- [x] Sensible stopping criterion* 

\* It should be avoided to compute `norm(b - Ax)` since that's as expensive as an iteration. Another option is `||x_next - x_curr|| / ||x_cur|| < tol`, but that might be complicated in the current Gauss-Seidel implementation as it updates `x` in-place. At the moment the stopping criterion is just the maximum number of iterations. That might be good enough as well: stationary methods are often used as smoothers in multigrid methods where just a small, fixed number of iterations is carried out.

Fixes #147


haampie:
At this point the sparse version on full matrices is significantly faster... :p

```
A = rand(30, 30) + 30I
b_sparse = @benchmark IterativeSolvers.jacobi!(x, mat, b, maxiter = 10) setup=(x=zeros(30); mat=sparse($A); b=$A*ones(30))
b_full = @benchmark IterativeSolvers.jacobi!(x, mat, b, maxiter = 10) setup=(x=zeros(30); mat=$A; b=$A*ones(30))
```

```
> b_sparse
BenchmarkTools.Trial:
  memory estimate:  1.05 KiB
  allocs estimate:  9
  minimum time:     11.584 μs (0.00% GC)
  samples:          10000
```

```
> b_full
BenchmarkTools.Trial:
  memory estimate:  126.45 KiB
  allocs estimate:  957
  minimum time:     71.540 μs (0.00% GC)
  samples:          10000
```

The current / full matrix version computes `norm(A*x-b)` which is allocating and as expensive as an iteration. But it might also have to do with row-wise matrix access.

haampie:
For dense matrices it makes sense to use the same column-wise approach as in the sparse implementation. See https://gist.github.com/haampie/6d24c91df406e9574f3d4cd451db2260

For matrices of size 300x300 you get a 2x speedup:

improvement

PR #158, opened 2017-08-11
Fix off-by-one error in iteration count of IDR(s)

haampie:
Fixes an off-by-one error in IDR(s). Current behaviour:

```julia
> length(idrs(rand(100, 100), rand(100), log=true, maxiter = 3)[2][:resnorm])
3
```

Previously:

```julia
> length(idrs(rand(100, 100), rand(100), log=true, maxiter = 3)[2][:resnorm])
ERROR: BoundsError: attempt to access 3-element Array{Float64,1} at index [4]
```


andreasnoack:
Could you add a test for this?

PR #162, opened 2017-08-24
Fix IterativeSolvers.jl for Julia 0.7

### LINKED ISSUE #161: invalid escape sequence in doc strings for lsqr and lmqr on Julia master
See https://travis-ci.org/JuliaMath/IterativeSolvers.jl/jobs/267798572#L2172

Docstring might have to use the `doc_str` macro.

cc: @haampie

haampie:
Fixes #161

### DIFF
--- a/src/svdl.jl
+++ b/src/svdl.jl
@@ -444,7 +444,9 @@ function harmonicrestart!(A, L::PartialFactorization{T,Tr},
     #Compute scaled residual from the harmonic Ritz problem
     r0 = zeros(Tr, m)
     r0[end] = 1
-    # isa(L.B, Bidiagonal) && @assert L.B.isupper
+    if isa(L.B, Bidiagonal)
+        @assert @static VERSION < v"0.7.0-DEV.884" ? L.B.isupper : L.B.uplo == 'U'
+    end
     r = try
         #(L.B\r0)
         A_ldiv_B!(L.B, r0)
--- a/src/svdl.jl
+++ b/src/svdl.jl
@@ -356,9 +356,8 @@ function build(log::ConvergenceHistory, A, q::AbstractVector{T}, k::Int) where {
     p = A*q
     α = convert(Tr, norm(p))
     @blas! p *= inv(α)
-    extend!(log, A, PartialFactorization(
-        reshape(p, m, 1), reshape(q, n, 1), Bidiagonal([α], Tr[], true), β
-        ), k)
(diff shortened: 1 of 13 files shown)


src/svdl.jl
```diff
@@ -356,9 +356,8 @@ function build(log::ConvergenceHistory, A, q::AbstractVector{T}, k::Int) where {
     p = A*q
     α = convert(Tr, norm(p))
     @blas! p *= inv(α)
-    extend!(log, A, PartialFactorization(
-        reshape(p, m, 1), reshape(q, n, 1), Bidiagonal([α], Tr[], true), β
-        ), k)
+    bidiag = Bidiagonal([α], Tr[], @static VERSION ≤ v"0.6" ? true : :U)
```
fredrikekre: This can be `@static VERSION < v"0.7.0-DEV.884" ? true : :U` to be more specific about the version.

haampie: Makes sense, thanks

haampie:
Unfortunately the tests require Plots.jl for the `test/history.jl` smoke test, and one of the dependencies fails on 0.7. 

I would really be in favor of just moving plotting tools to a different package like IterativeSolversTools.jl so that IterativeSolvers is very close to being self-contained.

test/history.jl
```diff
@@ -1,24 +0,0 @@
-using IterativeSolvers
-using Plots
```
andreasnoack: Is plotting no longer supported?

haampie: Well, the `test/history.jl` file is basically just calling `plot(...)` on the convergence history object of all kinds of solvers. It's not testing whether it's plotting the correct thing (there are no tests in that file).

Let me just remove this commit for now and open another PR later.

haampie: At any rate, the tests are failing on 0.7 now because of dependencies of Plots.jl, but we can't really help that.

src/svdl.jl
```diff
@@ -445,7 +444,7 @@ function harmonicrestart!(A, L::PartialFactorization{T,Tr},
     #Compute scaled residual from the harmonic Ritz problem
     r0 = zeros(Tr, m)
     r0[end] = 1
-    isa(L.B, Bidiagonal) && @assert L.B.isupper
+    # isa(L.B, Bidiagonal) && @assert L.B.isupper
```
andreasnoack: Better to make this check conditional on the version of Julia

PR #163, opened 2017-08-24
Fix element types

haampie:
This PR makes iterables less restrictive when it comes to vector types: `x` does not have to have the same type as `b`, and might not even have the same type as the residual vector `r` (for instance when `x` is view of a column in a matrix).

Also it uses `eltype(x)` over `eltype(b)` when it comes to initializing the residual vector or some temporary vector / matrix. This fixes a problem in Circuitscape.jl where `Ax=b` is solved with `b` an array of integers while the residual `r` should be a vector of floats.

Lastly some unused methods are removed from common.jl.


andreasnoack:
It would be great with some tests here where the rhs has integer elements or maybe even `BigInt`.

PR #171, opened 2017-12-02
Make maxiter the size of the matrix

### LINKED ISSUE #168: `maxiter` parameter
```
using IterativeSolvers
L1d(N) = spdiagm((-ones(N-1), 2*ones(N), -ones(N-1)), (-1,0,1), N, N) * N^2
function L2d(N)
  L1 = L1d(N)
  A = kron(speye(N), L1) + kron(L1, speye(N))
  b = ones(size(A,1))
  return A, b
end
A, b = L2d(100)
TOL = 1e-4
x = IterativeSolvers.cg(A, b; tol=1e-4)
norm(A * x - b, Inf)
```
The output of the final line is `35.59...` 

This is because `maxiter = size(A,2)` instead of `maxiter = size(A,2)^2` as I think it should be (and indeed as is claimed in the documentation).  I believe this must have changed recently. The above problem is just a finite-difference laplacian on a square domain. Something as basic as that should not fail with default parameters.

haampie:
Closes #168. cc @cortner

### DIFF
--- a/src/bicgstabl.jl
+++ b/src/bicgstabl.jl
@@ -28,7 +28,7 @@ bicgstabl_iterator(A, b, l; kwargs...) = bicgstabl_iterator!(zerox(A, b), A, b, 
 
 function bicgstabl_iterator!(x, A, b, l::Int = 2;
     Pl = Identity(),
-    max_mv_products = min(30, size(A, 1)),
+    max_mv_products = size(A, 2),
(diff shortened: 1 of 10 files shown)


andreasnoack:
Looks like the `DampenedMatrix` needs a `size` method.

cortner:
thank you for this

PR #174, opened 2017-12-06
Remove dependencies on UnicodePlots and LinearMaps

### LINKED ISSUE #172: Remove plotting from this package
I'm not sure what the general opinion is about this, but I feel the dependencies on `UnicodePlots` and `RecipesBase` are odd. 

Do we really need plotting? Considering you typically only plot the residual norm of each iteration, what's wrong with writing

```julia
using Plots
x, history = gmres(A, b, log = true)
plot(history[:resnorm], yscale = :log10)
```

Furthermore, plotting and `ConvergenceHistory` objects in general are not tested.

If we really want to have it, maybe it can be moved to another repository like IterativeSolversTools.jl as well.

haampie:
Closes #172.

UnicodePlots can be removed since there is a plot recipe already.

LinearMaps is only a test dependency.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,4 +2,3 @@ julia 0.6
 
 RecipesBase
 SugarBLAS
-LinearMaps
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,5 @@
 julia 0.6
 
-UnicodePlots
 RecipesBase
 SugarBLAS
 LinearMaps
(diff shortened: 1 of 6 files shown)


andreasnoack:
Is the plot recipe tested?

haampie:
Not yet. I think I see a way to test them by directly calling `RecipesBase.apply_recipe(..., ::ConvergenceHistory, ..)`

haampie:
There they are, finally :).

PR #184, opened 2017-12-25
Use broadcasting rather than BLAS calls

### LINKED ISSUE #181: Broadcasting rather than BLAS 1 calls?
There are some instances where BLAS calls can be replaced by broadcasting.

In CG and Chebyshev iteration there is an `axpby` which is currently written as `scale` and `axpy` (since `axpby` is not part of Julia 0.6). This looks wasteful:

```julia
scale!(cheb.u, β)
@blas! cheb.u += one(T) * cheb.c
```

could be replaced by

```julia
@. cheb.u = cheb.c + β * cheb.u
```

Is there any reason *not* to do this?  And while we're at it, can't we replace `axpy`'s with broadcasting in general?

cc'ing @andreasnoack because we probably discussed this earlier.

haampie:
Closes #181. Some code benefits here, like cg, chebyshev and bicgstabl. The only downside is that we're missing threading of openblas, but that helps only marginally on the CPU for BLAS 1.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,3 @@
 julia 0.6
 
 RecipesBase
-SugarBLAS
--- a/src/IterativeSolvers.jl
+++ b/src/IterativeSolvers.jl
@@ -5,8 +5,6 @@ eigensystems, and singular value problems.
 """
 module IterativeSolvers
 
-using SugarBLAS
-
 include("common.jl")
 include("orthogonalize.jl")
 include("history.jl")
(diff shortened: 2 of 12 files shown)


andreasnoack:
I still think this is fine but I realized one catch here. If you want to try out these methods with a new fancy array type then there is a tax from using dots since supporting broadcasting is much harder than supporting BLAS-1.

haampie:
True, I had to do it in the test of lsmr and it took me a while to figure out how to do it. Hopefully the docs are clear when Julia 1.0 is released

Issue #186, opened 2018-01-04
IDRS allocates on application of linear operator


samuelpowell:
The IDRS solver uses ```@inline linsys_op(x, A) = A*x``` to apply the supplied operator, resulting in allocation in the inner loop. It appears that this could be modified to use `A_mul_B!` without a significant refactor.

PR #187, opened 2018-01-04
Reduce allocation in IDRS inner loop

samuelpowell:
This reduces allocations in the IDRS inner loop, ref #186.

- I am unfamiliar with the algorithm and have thus made minimum changes.
- I am unsure of the reason for splatting the operator into the `linsys_op` function. Tests pass but perhaps I've misunderstood this implementation choice.

### DIFF
--- a/src/idrs.jl
+++ b/src/idrs.jl
@@ -45,7 +45,7 @@ function idrs!(x, A, b;
     history = ConvergenceHistory(partial=!log)
     history[:tol] = tol
     reserve!(history,:resnorm, maxiter)
-    idrs_method!(history, x, linsys_op, (A,), b, s, tol, maxiter; kwargs...)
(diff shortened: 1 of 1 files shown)


mschauer:
The design comes from a time where https://github.com/Jutho/LinearMaps.jl was not yet a thing and I wanted solve Lyapunov type equations. See https://github.com/mschauer/IDRsSolver.jl/blob/master/src/IDRsSolver.jl#L47 . Aligning the code with the other solvers is a good idea.

haampie:
Looks good to me! The splatting shouldn't be necessary when LinearMaps is used.

samuelpowell:
@mschauer ahh yes, that makes sense. Indeed, [LinearMaps.jl](https://github.com/Jutho/LinearMaps.jl) is incredibly useful for this kind of thing. 

@haampie thanks for review.

PR #193, opened 2018-03-24
Allow pre-allocation and re-use of buffers in CG

mohamed82008:


haampie:
Hey, I think this might be a good direction, since right now reusing the iterable structs is a bit awkward. This needs to be documented though!

src/cg.jl
```diff
@@ -90,13 +90,22 @@ end
 
 # Utility functions
 
+struct CGStateVariables{T}
+    u::Vector{T}
```
haampie: This is too restrictive. We should allow any vector type: SubArrays, SharedArrays etc.

haampie: Now that I'm thinking of it, in some cases `x` might be say a view of column of a matrix, while `u`, `r` and `c` are standard vectors. It's useful to allow this.

mohamed82008: Sure, will make it more general.

mohamed82008: Oh haven't noticed the second comment, I will make another change tomorrow.

mohamed82008: To allow the above case systematically across all matrix types, it would be a bit hard since we have to identify the most natural vector type to use which could be `Vector`, `GPUArray`, `SharedArray`, etc. With the exception of subarrays, the above should work out of box. And in case of subarrays, the user has the option of deciding on the buffer types, since I am using a very flexible type annotation `statevars::CGStateVariables`. Also this has the same behavior as before the commit, in fact now it is more flexible since the user can choose different types of `x` and the buffers should they have a good reason to.

mohamed82008:
I believe this is ready?

PR #196, opened 2018-04-08
Record resnorm in powm!

tkf:
as mentioned in the docstring.

### DIFF
--- a/src/simple.jl
+++ b/src/simple.jl
@@ -21,7 +21,7 @@ end
 
 @inline converged(p::PowerMethodIterable) = p.residual ≤ p.tol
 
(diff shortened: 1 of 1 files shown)


src/simple.jl
```diff
@@ -21,7 +21,7 @@ end
 
 @inline converged(p::PowerMethodIterable) = p.residual ≤ p.tol
 
-@inline start(p::PowerMethodIterable) = 0
+@inline start(p::PowerMethodIterable) = 1
```
tkf: Alternatively,  we can do `reserve!(history, :resnorm, maxiter + 1)` but iterating `maxiter + 1` instead of `maxiter` was a minor bug so why not fix it here.

PR #198, opened 2018-04-20
Update to Julia 0.7

lostella:
I noticed that there's little discussion on updating IterativeSolvers to Julia 0.7: since I started updating some packages depending on IterativeSolvers as well, I thought of giving it a shot.

For now it looks like errors are gone, but there's still quite a few warnings to get rid of. Will keep working on this, I'm opening the PR to start getting feedback in case something is not right or could be done better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,8 @@ language: julia
 os:
     - linux
 julia:
-    - 0.6
     - nightly
 matrix:
-  allow_failures:
-    - julia: nightly
 notifications:
     email: false
 after_success:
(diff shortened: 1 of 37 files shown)


mohamed82008:
Will this receive some love any time soon?

dkarrasch:
Just a few comments: I still seem to see a few `A_mul_B!` occurences, which should be replaced by `mul!`s altogether. In a currently reviewed PR in `LinearMaps`, the inplace matvec multiplication interface is also `LinearAlgebra.mul!`, and for not-inplace `Base.*`. In case the linear operator is given by a LinearMap `A`, a call like `A_mul_B!(y, A, x)` would not work.

lostella:
Closing in favour of #205 .

PR #200, opened 2018-06-04
LOBPCG solver

mohamed82008:
Implements the LOBPCG eigenvalue solver as per the Python implementation in https://github.com/scipy/scipy/blob/v1.1.0/scipy/sparse/linalg/eigen/lobpcg/lobpcg.py#L109-L568.

### DIFF
--- a/src/lobpcg.jl
+++ b/src/lobpcg.jl
@@ -286,7 +286,7 @@ function A_rdiv_B!(A, B::UpperTriangular)
     return A
 end
 
-realdiag!(M) = nothing
+realdiag!(M) = M
 function realdiag!(M::AbstractMatrix{TC}) where TC <: Complex
     @inbounds for i in 1:minimum(size(M))
         M[i,i] = real(M[i,i])
--- a/src/lobpcg.jl
+++ b/src/lobpcg.jl
@@ -281,7 +281,7 @@ function A_rdiv_B!(A, B::UpperTriangular)
         end
         A[:,i] .= view(A, :, i) ./ B[i,i]
     end
-    return
+    return A
 end
 
 realdiag!(M) = nothing
@@ -290,7 +290,7 @@ function realdiag!(M::AbstractMatrix{TC}) where TC <: Complex
     for i in 1:size(M, 1)
         M[i,i] = Complex(real(M[i,i]), zero(T))
     end
-    return
+    return M
 end
 
 function (ortho!::CholQR)(XBlocks::Blocks{Generalized}, sizeX = -1; update_AX=false, update_BX=false) where Generalized
--- a/src/lobpcg.jl
+++ b/src/lobpcg.jl
@@ -1,0 +1,793 @@
+#=
+The code below was derived from the scipy implementation of the LOBPCG algorithm in https://github.com/scipy/scipy/blob/v1.1.0/scipy/sparse/linalg/eigen/lobpcg/lobpcg.py#L109-L568. 
+
+Since the link above mentions the license is BSD license, the notice for the BSD license 2.0 is hereby given below giving credit to the authors of the Python implementation.
+
+Copyright (c) 2018, Robert Cimrman, Andrew Knyazev
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * The names of the contributors may not be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+=#
+
+export lobpcg, lobpcg!, LOBPCGIterator
+
+struct Blocks{Generalized, T, TA<:AbstractArray{T}}
+    block::TA # X, R or P
+    A_block::TA # AX, AR or AP
+    B_block::TA # BX, BR or BP
+end
+Blocks(X, AX) = Blocks{false, eltype(X), typeof(X)}(X, AX, X)
+Blocks(X, AX, BX) = Blocks{true, eltype(X), typeof(X)}(X, AX, BX)
+function A_mul_X!(b::Blocks, A)
+    A_mul_B!(b.A_block, A, b.block)
+    return
+end
+function A_mul_X!(b::Blocks, A, n)
+    A_mul_B!(view(b.A_block, :, 1:n), A, view(b.block, :, 1:n))
+    return
+end
+function B_mul_X!(b::Blocks{true}, B)
+    A_mul_B!(b.B_block, B, b.block)
+    return
+end
+function B_mul_X!(b::Blocks{true}, B, n)
+    A_mul_B!(view(b.B_block, :, 1:n), B, view(b.block, :, 1:n))
+    return
+end
+function B_mul_X!(b::Blocks{false}, B, n = 0)
+    return
+end
+
+struct Constraint{T, TA<:AbstractArray{T}, TC}
+    Y::TA
+    BY::TA
+    gram_chol::TC
+    gramYBV::TA # to be used in view
+    tmp::TA # to be used in view
+end
+function Constraint(::Void, B, X)
+    return Constraint{Void, Matrix{Void}, Void}(Matrix{Void}(0,0), Matrix{Void}(0,0), nothing, Matrix{Void}(0,0), Matrix{Void}(0,0))
+end
+function Constraint(Y, B, X)
+    T = eltype(X)
+    if B isa Void
+        B = Y
+    else
+        BY = similar(B)
+        A_mul_B!(BY, B, Y)
+    end
+    gramYBY = At_mul_B(Y, BY)
+    gramYBY_chol = cholfact!(Hermitian(gramYBY))
+    gramYBV = zeros(T, size(Y, 2), size(X, 2))
+    tmp = similar(gramYBV)
+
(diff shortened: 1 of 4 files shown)


mohamed82008:
All tests pass now.

mschauer:
We have to be a bit careful here, the BSD 3 clause licence has stonger requirements on reuse than MIT, therefore the entire package can not be under MIT if parts are under BSD 3.

mohamed82008:
This licensing issue was discussed on Slack, jsoc-standup channel, and the non-legally binding conclusion reached was that derived work can be used without changing the license of the package, so long as the license notice is added. One has to be careful when advertising or promoting the package though not to violate any of the BSD clauses.

Here is a good discussion https://opensource.stackexchange.com/questions/217/what-are-the-essential-differences-between-bsd-and-mit-licences.

mohamed82008:
I believe this PR is ready. @haampie

mohamed82008:
I can probably make the coverage 100% if I test constraints and preconditioners.

src/lobpcg.jl
```diff
@@ -0,0 +1,751 @@
...
+
+- `results`: a `LOBPCGResults` struct. `r.λ` and `r.X` store the eigenvalues and eigenvectors.
+
+"""
+function lobpcg!(iterator::LOBPCGIterator; log=false, tol=nothing, maxiter=200)
+    T = eltype(iterator.XBlocks.block)
+    X = iterator.XBlocks.block
+    for j in 1:size(X,2)
+        if all(x -> x==0, view(X, :, j))
+            X[:,j] .= rand.()
```
mschauer: I am curious, why is this working?

mohamed82008: This is to make sure we don't start with zeros in any of the Ritz vector. Any dependence in the vectors is going to kill the algorithm. I am assuming the random vectors will not be dependent.

src/lobpcg.jl
```diff
@@ -0,0 +1,751 @@
...
+- `maxiter`: maximum number of iterations; default is 200;
+
+- `tol::Number`: tolerance to which residual vector norms must be under.
+
+# Output
+
+- `results`: a `LOBPCGResults` struct. `r.λ` and `r.X` store the eigenvalues and eigenvectors.
+
+"""
+function lobpcg(A, largest::Bool, X0::Union{AbstractMatrix, AbstractVector}; kwargs...)
```
haampie: I think asking the user for an initial guess is a bit awkward; it's probably almost always a random vec or block. Default value would be useful.

src/lobpcg.jl
haampie: Can't we just drop this check? Seems wasteful to iterate over the data when it's generally not strictly necessary. Also `rand.()`?

src/lobpcg.jl
mohamed82008: Sure.

src/lobpcg.jl
```diff
@@ -223,14 +223,14 @@ function (g::BlockGram)(gram, lambda, n1::Int, n2::Int, n3::Int)
     if n2 > 0
         gram[rr, rr] .= view(g.RAR, 1:n2, 1:n2)
         gram[xr, rr] .= view(g.XAR, 1:n1, 1:n2)
-        transpose!(view(gram, rr, xr), view(g.XAR, 1:n1, 1:n2))
+        conj!(transpose!(view(gram, rr, xr), view(g.XAR, 1:n1, 1:n2)))
```
haampie: `adjoint!`?

mohamed82008: Not available in v0.6 it seems.

src/lobpcg.jl
mohamed82008: I am adding another wrapper function which defaults to `Matrix` and `Float64` and calls `size` on `A`. I think this is a nice compromise since the `X0` functions are still there.

mohamed82008: Oh `X0` also tells us the number of eigenvectors desired, so that will have to be replaced by `nev` or something..

src/lobpcg.jl
```diff
@@ -0,0 +1,751 @@
...
+        A[:,i] .= view(A, :, i) ./ B[i,i]
+    end
+    return
+end
+
+realdiag!(M) = nothing
+function realdiag!(M::AbstractMatrix{TC}) where TC <: Complex
+    T = real(eltype(M))
+    for i in 1:size(M, 1)
+        M[i,i] = Complex(real(M[i,i]), zero(T))
```
haampie: Let's keep things simple: `M[i,i] = real(M[i,i])` is equivalent

src/lobpcg.jl
```diff
@@ -0,0 +1,751 @@
...
+function A_rdiv_B!(A, B::UpperTriangular)
+    s = size(A, 2)
+    A[:,1] .= view(A, :, 1) ./ B[1,1]
+    for i in 2:s
+        for j in 1:i-1
+            A[:,i] .= view(A, :, i) - view(A, :, j) .* B[j,i]
+        end
+        A[:,i] .= view(A, :, i) ./ B[i,i]
+    end
+    return
```
haampie: Convention is to return `A`.

haampie: `@inbounds`? This is an O(n) operation right? Better make it fast -- although there's a blas wrapper for this on 0.7.

src/lobpcg.jl
```diff
@@ -0,0 +1,751 @@
...
+function (g::BlockGram)(gram, lambda, n1::Int, n2::Int, n3::Int)
+    xr = 1:n1
+    rr = n1+1:n1+n2
+    pr = n1+n2+1:n1+n2+n3
+    if n1 > 0
+        #gram[xr, xr] .= view(g.XAX, 1:n1, 1:n1)
+        gram[xr, xr] .= Diagonal(view(lambda, 1:n1))
+    end
+    if n2 > 0
+        gram[rr, rr] .= view(g.RAR, 1:n2, 1:n2)
```
haampie: When dotting you don't need views for the rhs, indexing is wrapped in `dotview`. Just putting the comment here, but applies everywhere.

src/lobpcg.jl
mohamed82008: This is a small price to pay. I think the most common pre-allocation function is `zeros` not `rand` so without this line, the function will fail more often than not when pre-allocating and the error will not be informative. I can add a flag that turns off this check for the trained user to use.

`rand.()` makes a random vector. `rand()` copies the same random number over all the elements.
```julia
julia> A = zeros(3);

julia> A .= rand()
3-element Array{Float64,1}:
 0.431159
 0.431159
 0.431159

julia> A .= rand.()
3-element Array{Float64,1}:
 0.740519
 0.313979
 0.199173
```

src/lobpcg.jl
haampie: We're advising to wrap the operator with LinearMaps.jl, which implements size and eltype for you.

src/lobpcg.jl
mohamed82008: Is this in v0.6?
```julia
julia> A = zeros(300000);

julia> @time A[1:end] .= A[1:end];
  0.003292 seconds (332 allocations: 2.308 MiB)

julia> @time A[1:end] .= A[1:end];
  0.001019 seconds (32 allocations: 2.290 MiB)

julia> @time A[1:end] .= view(A, 1:length(A));
  0.000241 seconds (47 allocations: 1.406 KiB)

julia> @time A[1:end] .= view(A, 1:length(A));
  0.000181 seconds (46 allocations: 1.359 KiB)
```

src/lobpcg.jl
mohamed82008: Sure.

src/lobpcg.jl
mohamed82008: Ok

src/lobpcg.jl
mohamed82008: Ok

src/lobpcg.jl
haampie: Ah, I was confused! Only the lhs is wrapped in dotview. For some heavy `view` sections you could also use the `@views` macro which wraps indexing ops in `maybeview`, which might be slightly cleaner -- but that's for you to judge.

src/lobpcg.jl
```diff
@@ -288,7 +288,7 @@ end
 
 realdiag!(M) = nothing
```
haampie: Last detail: `nothing` -> `M`

src/lobpcg.jl
haampie: :heavy_check_mark:

src/lobpcg.jl
mohamed82008: I will stick with the current one then. It's less appealing but more informative about what's happening.

mohamed82008:
I will give this one more cleanup and testing wave to increase coverage as much as possible. There seems to be some unused code and other untested code.

src/lobpcg.jl
```diff
@@ -200,9 +201,9 @@ PAP!(BlockGram, PBlocks, n) = Ac_mul_B!(view(BlockGram.PAP, 1:n, 1:n), view(PBlo
 XBP!(BlockGram, XBlocks, PBlocks, n) = Ac_mul_B!(view(BlockGram.XAP, :, 1:n), XBlocks.block, view(PBlocks.B_block, :, 1:n))
 XBR!(BlockGram, XBlocks, RBlocks, n) = Ac_mul_B!(view(BlockGram.XAR, :, 1:n), XBlocks.block, view(RBlocks.B_block, :, 1:n))
 RBP!(BlockGram, RBlocks, PBlocks, n) = Ac_mul_B!(view(BlockGram.RAP, 1:n, 1:n), view(RBlocks.B_block, :, 1:n), view(PBlocks.block, :, 1:n))
-XBX!(BlockGram, XBlocks) = Ac_mul_B!(BlockGram.XAX, XBlocks.block, XBlocks.B_block)
-RBR!(BlockGram, RBlocks, n) = Ac_mul_B!(view(BlockGram.RAR, 1:n, 1:n), view(RBlocks.block, :, 1:n), view(RBlocks.B_block, :, 1:n))
-PBP!(BlockGram, PBlocks, n) = Ac_mul_B!(view(BlockGram.PAP, 1:n, 1:n), view(PBlocks.block, :, 1:n), view(PBlocks.B_block, :, 1:n))
+#XBX!(BlockGram, XBlocks) = Ac_mul_B!(BlockGram.XAX, XBlocks.block, XBlocks.B_block)
+#RBR!(BlockGram, RBlocks, n) = Ac_mul_B!(view(BlockGram.RAR, 1:n, 1:n), view(RBlocks.block, :, 1:n), view(RBlocks.B_block, :, 1:n))
+#PBP!(BlockGram, PBlocks, n) = Ac_mul_B!(view(BlockGram.PAP, 1:n, 1:n), view(PBlocks.block, :, 1:n), view(PBlocks.B_block, :, 1:n))
```
mschauer: Out-commented code

mohamed82008: Yes it is unused since this stuff is normalized. I forgot to add that to the commit message.

mohamed82008: It was used for sanity checks and it might be useful when improving the package in the future so I did not remove it entirely. I am just trying to increase coverage!

mohamed82008:
With the exception of `show` functions, all the code is now covered by tests.

haampie:
Let's think a bit more about API design and "constraints". I'm not familiar with the term, since it's usually "deflation", and maybe users are not either?  In principle it comes down to applying lobpcg with the matrix `(I - VV')(A - θBI)(I - VV')` right? And the preconditioner would be applied as `(I - VV')inv(P)(I - VV')`.

What about an API similar to `eigs` where we can specify how many eigenvalues we want. If the block size happens to be smaller than the number of wanted eigenvalues, we can always add these converged guys to a "constraint" block, right? Or should the block size always be larger than the number of eigenvalues you want? 

Also, is there any benefit in copying converged eigenvectors from the block over to some constraint matrix? Do you observe a situation where part of the eigenvectors in a block are converged and the other part is not? Can we avoid redundant mv-products in that case?

mohamed82008:
Is this a feature request, decoupling `nev` and block size that is? I could do that, it would just involve adding another API function.

mohamed82008:
I tried decoupling them and one subtle point/problem I found was that it is not clear what to do when one block does not converge within the limit on the number of iterations. I added the current Ritz vectors to the constraints anyways, but the correctness of that is questionable.

mohamed82008:
Pretty good coverage at this point. All tests pass. I hope this gets merged this soon, I really want to move on!

haampie:
I think you can move on in the meantime ^^

Issue #191, opened 2018-02-04
Julia 0.7


haampie:
Besides the obvious things there is some room for improvement:

- [ ] Use the new iterator interface https://github.com/JuliaLang/julia/pull/25261 everywhere
- [ ] Use new logging goodies
- [ ] ...

mohamed82008:
Close issue?

PR #205, opened 2018-07-07
Julia 0.7 update

lostella:
Addressing #191. This is an updated re-issuing of #198, taking into account all recent contributions to IterativeSolvers and changes to Julia 0.7. It relies on [this PR for LinearMaps](https://github.com/Jutho/LinearMaps.jl/pull/23), so CI will fail before that is merged.

Open issues:
- [x] 3 occurrences of `spdiagm` (1 in benchmark, 2 in test) still need be updated and raise warnings as of now (I'm looking into this now).
- [x] I temporarily commented "Dampened tests" for LSMR, apparently the utility code for `DampenedVector` and `DampenedMatrix` needs to be extended to have this test working; I'm happy to fix that with some help from the authors.
- [x] Some tests (lobpcg.jl and stationary.jl) worked after changing the rng seed, otherwise some assertions complained about tolerances not being met (by a small margin) or matrices not being positive semi-definite.
    - [x] stationary.jl
    - [x] lobpcg.jl

I'd like to get inputs especially on the latter two items, as well as on the rest of the code of course.
For the rest, this PR should be merged as soon as [this PR for LinearMaps](https://github.com/Jutho/LinearMaps.jl/pull/23) is.

**Edit:** since this PR is 0.6-incompatible, I would also immediately
- [x] Update iterables interface.

### DIFF
--- a/src/lobpcg.jl
+++ b/src/lobpcg.jl
@@ -1,5 +1,5 @@
 #=
-The code below was derived from the scipy implementation of the LOBPCG algorithm in https://github.com/scipy/scipy/blob/v1.1.0/scipy/sparse/linalg/eigen/lobpcg/lobpcg.py#L109-L568. 
+The code below was derived from the scipy implementation of the LOBPCG algorithm in https://github.com/scipy/scipy/blob/v1.1.0/scipy/sparse/linalg/eigen/lobpcg/lobpcg.py#L109-L568.
 
 Since the link above mentions the license is BSD license, the notice for the BSD license 2.0 is hereby given below giving credit to the authors of the Python implementation.
 
@@ -29,6 +29,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 =#
 
 export lobpcg, lobpcg!, LOBPCGIterator
+
+using LinearAlgebra
+using Random
 
 struct LOBPCGState{TR,TL}
     iteration::Int
@@ -62,8 +65,8 @@ struct LOBPCGResults{TL, TX, T, TR, TI, TM, TB, TTrace <: Union{LOBPCGTrace, Abs
 end
 function EmptyLOBPCGResults(X::TX, k::Integer, tolerance, maxiter) where {T, TX<:AbstractMatrix{T}}
     blocksize = size(X,2)
-    λ = Vector{T}(k)
-    X = TX(size(X, 1), k)
+    λ = Vector{T}(undef, k)
+    X = TX(undef, size(X, 1), k)
 
     iterations = zeros(Int, ceil(Int, k/blocksize))
     residual_norms = copy(λ)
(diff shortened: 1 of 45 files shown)


dkarrasch:
Regarding the norm issue, a quick check indeed seems to say that only norms of vectors are computed, which is to be expected in the context of iterative solvers. At times, however, this is done via `vecnorm` (I checked the current version, not the PR), which is deprecated in favor of `norm`.

lostella:
I added somewhat minimal changes to make the package compliant with the new iteration protocol. These can easily be extended by completely getting rid of `start` and `done`.

Further restructuring & uniformization of the iterators I would defer to a later stage.

haampie:
Thanks, this looks great, I'll look into that issue with stationary.jl. The DampenedVector stuff is imho a bit too much for a test, but let's see if it can be upgraded easily as well.

haampie:
Hm, I think I'll just simplify the DampenedVector bit, cause as far as I see the only special thing is the DampenedMatrix.

lostella:
Yeah it definitely looks like there's something fishy with test tolerances in lobpcg.jl.

src/lobpcg.jl
```diff
@@ -869,9 +872,9 @@ function lobpcg!(iterator::LOBPCGIterator; log=false, tol=nothing, maxiter=200,
     end
     n = size(X, 1)
     sizeX = size(X, 2)
-    residualTolerance = (tol isa Void) ? (eps(real(T)))^(real(T)(4)/10) : real(tol)
+    residualTolerance = (tol isa Nothing) ? (eps(real(T)))^(real(T)(4)/10) : real(tol)
```
lostella: Could the problem be here? @mohamed82008 
Tests may fail since
```
(eps(real(T)))^(real(T)(4)/10) > sqrt(eps(real(T)))
```
but the test tolerance is set to `sqrt(eps(real(T)))`.

lostella: Nevermind, `tol` is actually not a `Nothing` in tests: it is passed in and matches the test tolerance.

mohamed82008: LOBPCG tests pass on my machine.

mohamed82008: Oh I see you changed the seed.

lostella: It also runs fine on my machine (yes, with a modified seed), but apparently not on Travis.

mohamed82008:
All good now. It seems like LOBPCG works, but some other tests fail on my machine and there are other warnings. I will try to fix the warnings.

lostella:
Yep, thank you @mohamed82008! However, I don't see any warning on macos, and Travis doesn't show any on Linux.

Coverage decrease may be due to the `@test_skip`ped tests: these will be re-enabled as soon as the PR for LinearMaps is merged (& version-bumped?)

mohamed82008:
Yes the warnings are gone when I went to nightly, seems like it was some bug in the early v0.7 beta.

clason:
Out of interest: I notice that https://github.com/JuliaRegistries/Uncurated / https://github.com/JuliaLang/METADATA.jl aren't actually picking up the new version (nor 0.6.0) -- is it by design that these two versions aren't tagged and published there?

haampie:
@clason thanks for reminding me; there was an issue with the tests when tagging IterativeSolvers.jl on METADATA.jl. @mohamed82008 http://juliarun-ci.s3.amazonaws.com/1789ca0b09724ec0b42d7f890223157a2e4900eb/pull_request_of_IterativeSolvers_on_julia_0_6.log this is probably resolved with the fixes of this PR? If so I'll check if I can redo the 0.6.0-tag with these changes and then publish again.

Issue #211, opened 2018-07-12
Docs for LOBPCG method missing


dkarrasch:
Scanning through the releases I realized that at version v0.6.0 another eigensolver named LOBPCG was introduced. Text searching the docs didn't yield a single hit. Would be nice to have a little bit of background and recommendations for usage etc.

mohamed82008:
That's unfortunate. Is there any specific reason why Documenter.jl (or Literate.jl) was not used before to put the docstrings into the docs, just time limitation? I never used these before but I am sure you know they are quite popular, especially Documenter. I think Literate.jl can also add Jupyter notebooks to the docs.

mohamed82008:
I think JuMP uses http://www.sphinx-doc.org/en/master/

haampie:
Just add a page to https://github.com/JuliaMath/IterativeSolvers.jl/tree/master/docs/src and reference it in https://github.com/JuliaMath/IterativeSolvers.jl/blob/master/docs/make.jl, that's all.

mohamed82008:
Close issue?
https://juliamath.github.io/IterativeSolvers.jl/latest/eigenproblems/lobpcg.html

PR #213, opened 2018-07-17
Add LOBPCG docs and bump docs/make.jl Julia version to v0.7

mohamed82008:
Please check if this is correct, I am not sure how to test it.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.7
+julia 0.7-beta
 
 RecipesBase 0.3.1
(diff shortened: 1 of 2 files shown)


haampie:
Fixes #211

haampie:
> It also admits a preconditioner and a "constraints" matrix C, such that the algorithm returns the smallest (or largest) eigenvalues associated with the eigenvectors in the nullspace of C'B.

This does not really make sense to me

mohamed82008:
The nullspace of A is all vectors x such that Ax = 0. This means that x is orthogonal to all the rows of A. If x is in the nullspace of C'B, then Bx is orthogonal to the rows of C', which are the columns of C. Or in other words, x is B-orthogonal to the columns of C. Long winded but correct :)

haampie:
Yeah, sure, but the wording is not really what I'd expect in this context. You'd probably want to call it deflation. Also it leaves some questions open: Is C supposed to be B-orthogonal? In that case it really is deflation with (I - CC'B) the projection operator in the standard inner product.

PR #229, opened 2018-11-15
Project.toml and Manifest.toml files

mohdibntarek:
This PR ~~replaces the REQUIRE files with~~ adds Project.toml and Manifest.toml files. This also fixes the test failures in https://github.com/JuliaMath/IterativeSolvers.jl/pull/227 and https://github.com/JuliaMath/IterativeSolvers.jl/pull/228 due to the breaking changes in the latest release of RecipesBase v0.6.0.


fredrikekre:
If you plan on registering new versions of the package you need `REQUIRE` still.

PR #230, opened 2018-11-30
Fix vector types in cg state

baggepinnen:
This solves this issue https://github.com/kul-forbes/ProximalOperators.jl/issues/64
When `x` is a `SubArray`, `zero(x)` creates an `Array` and the types will not be the same. `similar(x::SubArray)` also results in an `Array`


haampie:
Thanks for pointing this out! :) I think a slightly more idiomatic way is to remove the type definitions and make Julia figure it out. I think this is what @mohamed82008 also did in #227.

haampie:
Closing this in favour of #231. Thanks for pointing this out and opening a pr!

PR #233, opened 2018-12-28
Fix DiagonalIndices when sparse matrix indices are not Int64.

tkonolige:
`searchsortedfirst` (used to get diagonal entries) requires its inputs to be Int64. When the sparse matrix passed to DiagonalIndices has a different type of indices, `searchsortedfirst` is not found. I've fixed this issue and added tests to cover this behavior.

### DIFF
--- a/src/stationary_sparse.jl
+++ b/src/stationary_sparse.jl
@@ -12,8 +12,8 @@ struct DiagonalIndices{Tv, Ti <: Integer}
         diag = Vector{Ti}(undef, A.n)
 
         for col = 1 : A.n
-            r1 = A.colptr[col]
-            r2 = A.colptr[col + 1] - 1
+            r1 = Int64(A.colptr[col])
+            r2 = Int64(A.colptr[col + 1] - 1)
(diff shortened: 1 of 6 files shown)


src/stationary_sparse.jl
```diff
@@ -12,8 +12,8 @@ struct DiagonalIndices{Tv, Ti <: Integer}
         diag = Vector{Ti}(undef, A.n)
 
         for col = 1 : A.n
-            r1 = A.colptr[col]
-            r2 = A.colptr[col + 1] - 1
+            r1 = Int64(A.colptr[col])
```
haampie: This should be `Int`, not `Int64`

haampie:
Thanks for the observation and PR! :)

This seems like an overly restrictive signature in Base, so you might want to file an issue in https://github.com/JuliaLang/julia

Further it's best to work with `Int` (not `Int64`) since that picks the native integer type of the platform (either Int32 or Int64).

haampie:
Thanks! New version is pending: https://github.com/JuliaLang/METADATA.jl/pull/20256

PR #238, opened 2019-01-15
Return CGResult from cg and allow absolute tol

### LINKED ISSUE #236: CGResult that saves info
It would be nice to have a `CGResult` struct that saves things like the last residual, number of iterations needed to converge, and options. If this is approved in principle, I can make a PR. This will be a breaking change however. A similar thing may also need to be done for the other algorithms but I am mostly interested in CG now.

mohdibntarek:
This fixes #236. I also allowed passing absolute tolerance as opposed to just a relative one. Since it makes sense for the residual to be <= the tolerance, the absolute tolerance is the one that should be returned in the `CGResult`. And it would be weird if I pass `tol = ..` to `cg` and then I see a different tolerance printed in the result.

### DIFF
--- a/test/cg.jl
+++ b/test/cg.jl
@@ -26,16 +26,17 @@ Random.seed!(1234321)
         A = rand(T, n, n)
         A = A' * A + I
         b = rand(T, n)
-        tol = √eps(real(T))
+        reltol = √eps(real(T))
 
-        r = cg(A, b; tol=tol, maxiter=2n, log=true)
+        r = cg(A, b; reltol=reltol, maxiter=2n, log=true)
         x, ch = r.x, r.history
         @test isa(ch, ConvergenceHistory)
-        @test norm(A*x - b) / norm(b) ≤ tol
+        @test norm(A*x - b) / norm(b) ≤ reltol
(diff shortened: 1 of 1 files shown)


mohamed82008:
I believe this is ready for a first review.

pkofod:
How do you plan to deprecate tol? I think it'd be better to add an `abs_tol` and `rel_tol`, have `tol` default to `nothing` and throw a deprecation warning if `tol` is actually set. 

(I came here because I wanted to use abs_tol in gmres for a inexact newton method, so I'd love if all methods got this abs_/rel_tol separation. Isn't base using abstol and reltol btw?)

mohamed82008:
I don't mind getting rid of `tol`. Not big on the underscore either.

pkofod:
Good. Then I think introducing `abstol` and `reltol` and deprecating the `tol` keyword is the best way forward.

chriscoey:
bump

ranocha:
I'm working on introducing keyword arguments `abstol` and `reltol` in #280 with a proper deprecation of `tol`. Once that's settled, this approach to change the return type would be nice to consider next.

mschauer:
#280 is merged now

PR #244, opened 2019-04-15
CG's stopping criteria

nrontsis:
# The problem
CG does not respect the stopping criterion of the documentation: 

> |r_k| / |r_0| ≤ tol

as it can be observed from the following minimal example:
```
using IterativeSolvers

n = 3
A = Matrix(I, n, n)
b = randn(n)

x = b + 1e-5*randn(n) # Initialize x close to the solution b
tol = 1e-3
initial_residual = norm(A*x - b)

output, history = cg!(x, A, b, tol=tol, initially_zero=false, log=true, verbose=true)
@assert norm(A*x - b)/initial_residual <= tol
```

It appears that `cg` stops when `|r_k| / |b| ≤ tol`. However, I am unsure if this actually holds, as the `CGStateVariables` are filled with vectors generated with the `similar()` command which can be problematic, as it does not initialise the memory of the vectors.

# Suggested solution

Change the stopping criterion to match the one of the documentation and replace `similar()` with `zero()`.

I suppose that another possibility would be to change the documentation? Let me know what you think!

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -128,7 +128,7 @@ function cg_iterator!(x, A, b, Pl = Identity();
     # Compute r with an MV-product or not.
     if initially_zero
         mv_products = 0
-        c = zero(x)
+        c .= zero(eltype(c))
         residual = norm(b)
         reltol = residual * tol # Save one dot product
     else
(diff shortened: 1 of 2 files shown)


chriscoey:
bump!

ranocha:
Bump. That's a good observation! See also #280.

PR #253, opened 2019-08-08
fix idrs! when near solution

### LINKED ISSUE #222: Error when initializing idrs! from [very near] solution
When initializing `idrs!` from near to the solution, I receive an error that res is not defined.  In looking closer, this is likely because `res` is referenced only on line 88 of idrs.jl and is not created anywhere. Line 88 is only called if `normR<=tol`, which is commented to occur when the "initial guess is a good enough solution."  Replacing the return value in this case to `X, log` seems to work fine for me, though this may be bad/breaking for other reasons. I thought I would report this here in case this or another solution would be an easy fix.  Thank you for putting this package together!

platawiec:
Fixes #222

### DIFF
--- a/src/idrs.jl
+++ b/src/idrs.jl
@@ -84,8 +84,8 @@ function idrs_method!(log::ConvergenceHistory, X, A, C::T,
         T_s = zero(R)
     end
 
-    if normR <= tol           # Initial guess is a good enough solution
-        return X, ConvergenceHistory(0<= res[end] < tol, tol, length(res), res)
+    if normR <= tol # Initial guess is a good enough solution
+        return X
     end
 
     Z = zero(C)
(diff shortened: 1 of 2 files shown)


mschauer:
Looks good, even with a Test! Is any of the bookkeeping 
```
            shrink!(log)
	    setconv(log, 0<=normR<tol)
	    setmvps(log, iter)
```
needed/useful for this early break?

platawiec:
Good point, I had a look. Only `setconv` seems necessary In fact, there was an extraneous call to `shrink!` which I removed.

chriscoey:
bump

mschauer:
bump?

mschauer:
Planning to merge after the weekend.

PR #255, opened 2019-08-25
QMR

platawiec:
I've finished a bare-bones QMR implementation. No pre-conditioning or look-ahead Lanczos (yet). Please have a look and let me know what you think.

I've mostly followed the style for MINRES. I split the Lanczos bi-orthogonalization process from the QMR iteration, in keeping with how the algorithms are abstracted in the original papers.

This algorithm should work with GPUs though it remains untested.


mschauer:
Thank you. How would you assess this code yourself? Do you think it is good to go once checked?

mschauer:
Thank you, this is really helpful

platawiec:
Can this be merged? Are there any open items left to address?

Issue #251, opened 2019-07-11
GPU friendly BLAS calls


shashi:
@ChrisRackauckas found out that using CuArrays fails at points like https://github.com/JuliaMath/IterativeSolvers.jl/blob/17ef2613dff1d80724886797160ee01bd1bd2c71/src/simple.jl#L38

where the call to `BLAS.axpy!` explicitly assumes that the inputs are OK with OpenBLAS. The right alternative would be to use `LinearAlgebra.axpy!` (which as it turns out is a different function) and let it naturally dispatch to BLAS iff possible.

pkofod:
see also #245

simonbyrne:
There are also several hard-coded `Matrix` types, e.g.
https://github.com/JuliaMath/IterativeSolvers.jl/blob/17ef2613dff1d80724886797160ee01bd1bd2c71/src/gmres.jl#L7

pkofod:
Since I need this, I might be able to go over some of these. If anyone has a usecase/ something that doens't currently work, please post it so I can take it into account.

simonbyrne:
We have some of our own code implemented:
https://github.com/climate-machine/CLIMA/tree/master/src/LinearSolvers
feel free to take/adapt what you need.

See also this comment for our needs: https://github.com/climate-machine/CLIMA/pull/330#issuecomment-516100717

Issue #245, opened 2019-05-02
Generic methods for BiCGStab and GMRES


spacedome:
Currently the implementations of BiCGStab and GMRES use methods in LinearAlgebra that do not have generic fallbacks. 

BiCGStab makes an explicit call to `BLAS.gemv!` which is implemented only for real and complex floats. I know that the LinearAlgebra package tries to have generic fallback methods, so I am not sure if this is an issue of IterativeSolvers, but currently the code does not work unless I implement `gemv!` for whatever type I'm using. 

GMRES also calls `BLAS.gemv!` but even if this is fixed, when computing the FastHessenberg the LinearAlgebra function `givensAlgorithm` is called, which is only defined for real and complex. Fixing the `gemv!` means GMRES would work with BigFloats and other real and complex types, but I am not sure how generic Givens rotations could be implimented.

Here is an example using Quaternions.jl 
```
using IterativeSolvers
using Quaternions
import Random: AbstractRNG, SamplerType

# rand not implemented in Quaternions, needed to run, but ignore this 
Base.rand(r::AbstractRNG, ::SamplerType{Quaternion{T}}) where {T<:Real} =
    Quaternion(rand(r,T), rand(r,T), rand(r,T), rand(r,T))

q = quat(1.0, 1.0, 1.0, 1.0)
A = [q q'; q' q]
b = [quat(1.0), quat(1.0)]
bicgstabl(A, b, 2)
```
 this gives the following error:
```
MethodError: no method matching gemv!(::Char, ::Quaternion{Float64}, ::SubArray{Quaternion{Float64},2,Array{Quaternion{Float64},2},Tuple{Base.Slice{Base.OneTo{Int64}},UnitRange{Int64}},true}, ::Array{Quaternion{Float64},1}, ::Quaternion{Float64}, ::SubArray{Quaternion{Float64},1,Array{Quaternion{Float64},2},Tuple{Base.Slice{Base.OneTo{Int64}},Int64},true})
```
The same error occurs with BigFloats.

Should LinearAlgebra have a generic fallback for `gemv!`? Should this package have a fallback for calls to `gemv!` based on type? I'm not entirely sure of the best way to do the latter, but it shouldn't be difficult, I can try to make a pull request if wanted. For GMRES, I'm not aware of a way to generalize Givens rotations, so maybe it should be explicit about only accepting real and complex? Not sure how important generic implimentations are to other people using this package.

chriscoey:
[...]

PR #261, opened 2019-12-20
WIP: Replaced BLAS calls with generic equivalents from LinearAlgebra

jbmuir:
This answers the letter but not the spirit of #251 and to an extent #245; there remain instances where matrices are created on the cpu (e.g. creating the ArnoldiDecomp for gmres) which remain to be resolved. This is my first ever PR so please let me know if I've stuffed something up!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.5'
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
>there remain instances where matrices are created on the cpu (e.g. creating the ArnoldiDecomp for gmres) 

Might need to use Adapt.jl

PR #263, opened 2020-01-26
adding type to reserve! in most solvers

nestordemeure:
When storing a log, solvers reserve memory with `reserve!`. By default no type information is passed to this method which, then, converts the history into Float64.
This is not the expected behaviour when using non traditional types (for instrumentation purpose) and can lead to problems if their silent convertion into Float64 is not detected by the programmer.

Thus I replaced several occurences of `reserve!(history, ...` with `reserve!(typeof(tol), history, ...`.

I did not modify `lsqr.jl`, `lsmr.jl` and `svdl.jl` as the modification was less straightforward but they should probably be modified in the same way.

(overall I think that `reserve!` should always stipulate the expected type to avoid regression on this issue)

### DIFF
--- a/src/bicgstabl.jl
+++ b/src/bicgstabl.jl
@@ -185,7 +185,7 @@ function bicgstabl!(x, A, b, l = 2;
     history[:tol] = tol
 
     # This doesn't yet make sense: the number of iters is smaller.
-    log && reserve!(history, :resnorm, max_mv_products)
+    log && reserve!(typeof(tol), history, :resnorm, max_mv_products)
 
     # Actually perform CG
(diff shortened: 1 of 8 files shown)


mschauer:
Out of curiosity, how did you run into this?

nestordemeure:
I did tests with an instrumented type that does not implement silent conversion to Float64 to avoid bugs due to unseen conversions.

ranocha:
Bump. This looks like a fix that would be nice to have in IterativeSolvers.jl.

ranocha:
What do you think of this, @haampie?

PR #266, opened 2020-02-12
This comma does not belong here.

mschauer:
It's not exactly clear to me how this did run all those years and apparently worked fine.

Edit: Apparently it worked as a nested loop, iterating over the result of the dot product and using `f[i]` as iteration variable and working this way.


mschauer:
Merging in 24h.

PR #268, opened 2020-03-10
Fix division by zero due to exact off-diagonal zeros in Hessenberg matrix

### LINKED ISSUE #267: gmres: nans with complex
```
using LinearMaps
using IterativeSolvers

T = ComplexF64
b = randn(T, 2)
Jop(x) = 2x
J = LinearMap{T}(Jop, 2)
x = gmres(J, b, verbose=true)
```
sometimes gets me NaNs

haampie:
Closes #267

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IterativeSolvers"
 uuid = "42fd0dbc-a981-5370-80f2-aaf504508153"
-version = "0.8.0"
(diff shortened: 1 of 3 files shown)


haampie:
@mschauer seems like the patch version in project.toml was not the same as the latest tag, so I'm bumping it to 0.8.2 here then and hope tag bot fixes everything for us :D

PR #274, opened 2020-05-01
Add import statement to make example runnable

GregVernon:
Currently, this example fails on the first line as `sprand` and `I` haven't been imported.  I discussed this in this Discourse comment: https://discourse.julialang.org/t/non-friendly-documentation/38109/112


ranocha:
Thanks, @GregVernon! I fixed the import statements and added the relevant part to the doctests in #285. That PR will supersede this PR.

PR #280, opened 2020-12-07
Absolute and relative tolerance for linear solvers

### LINKED ISSUE #207: add abs_tol as termination criterion
Sometimes the absolute tolerance is more useful than the relative tolerance in case of newton iteration. Could someone add it as a termination criteron?

ranocha:
This PR adds keyword arguments `abstol` and `reltol` to iterative solvers for linear system and deprecates the former keyword `tol` (which is now `reltol`).

- [x] BiCGStab(l)
- [x] CG
- [x] Chebyshev
- [x] GMRES
- [x] IDR(s)
- [x] MINRES
- [x] QMR
- LSMR doesn't seem to fit to this common framework
- LSQR doesn't seem to fit to this common framework

Since this is my first contribution to IterativeSolvers.jl, I would like to get some feedback on this work to improve this PR. Thanks!

If this approach is acceptable, I will continue to work on the other iterative solvers and adapt them accordingly.

By the way, I noticed that CG doesn't respect the stopping criterion described in the docs, see also #244. I think it would be good to fix this (as described in my comment in `cg.jl`).

Closes #207 by adding the keyword argument `abstol`.

Closes #244 by fixing the convergence criterion for CG (`reltol` should be relative to the initial residual instead of the RHS `b`).


src/cg.jl
```diff
@@ -114,40 +118,48 @@ struct CGStateVariables{T,Tx<:AbstractArray{T}}
...
         r .-= c
-        residual = norm(r)
-        reltol = norm(b) * tol
     end
+    residual = norm(r)
+    # TODO: According to the docs, the code below should use the initial residual
+    #       instead of the norm of the RHS `b` to set the relative tolerance.
+    # See also https://github.com/JuliaMath/IterativeSolvers.jl/pull/244
+    # tolerance = max(reltol * residual, abstol)
+    tolerance = max(reltol * norm(b), abstol)
```
haampie, line 152: Just use the initial residual, it saves another inner product.

ranocha: Okay, thanks, I will do that and change the behavior to be in accordance with the docs. I will have to change some tests, too.

ranocha: I've made this change. Based on this comment, I'll assume that I shall do the same for the other iterative methods.

ranocha:
I made all the changes to the iterative solvers for "simple" linear problems. Since the solvers for least-squares problems use a different setup of tolerances, I didn't adapt them. It would be nice to have another round of reviews to improve this PR :slightly_smiling_face:

test/idrs.jl
```diff
@@ -57,4 +58,30 @@ end
...
+        # If the initial residual is small and a small relative tolerance is used,
+        # many iterations are necessary
+        x = x0 + sqrt(eps(real(T))) * perturbation
+        initial_residual = norm(A * x - b)
+        x, ch = idrs!(x, A, b, log=true)
+        @test 2 ≤ niters(ch) ≤ n
+
+        # If the initial residual is small and a large absolute tolerance is used,
+        # no iterations are necessary
+        x = x0 + 10*sqrt(eps(real(T))) * perturbation
```
mschauer, line 80: The comment seem to be at odds with the larger initial residual?

ranocha: I'm sorry, I don't understand what you mean, @mschauer. The initial residual `A * x - b` is small here since `x = A \ b`.

test/idrs.jl
```diff
@@ -57,4 +58,30 @@ end
...
+              -1  2 -1
+               0 -1  2]
+        n = size(A, 2)
+        b = ones(T, n)
+        x0 = A \ b
+        perturbation = T[(-1)^i for i in 1:n]
+
+        # If the initial residual is small and a small relative tolerance is used,
+        # many iterations are necessary
+        x = x0 + sqrt(eps(real(T))) * perturbation
```
mschauer, line 73: here you have a different initial residual size and you write “small” in both cases. Just something I stumbled upon

ranocha: Ah, I see. Thanks for noticing and explaining this. I modified these tests such that the same initial residual is used for the tests of the absolute and relative tolerances (0869558).

mschauer:
Can one express the logic/stopping criteria with `reltol` and `abstol`  using `isapprox` from base?

ranocha:
> Can one express the logic/stopping criteria with `reltol` and `abstol` using `isapprox` from base?

`isapprox(x, y)` returns `true` if 
```julia
norm(x-y) <= max(atol, rtol*max(norm(x), norm(y)))
```
Our stopping criterion is 
```julia
norm(residual) <= max(abstol, reltol * initial_residual_norm)
```
Hence, I don't see a possibility to do that.

mschauer:
I see, thanks.

ranocha:
@haampie Can we merge this? The decreased coverage is only caused by the new depwarns.

haampie:
Hi @ranocha , I haven't had the chance to review yet. Tomorrow I can do that

haampie (approved):

haampie:
Thank you!

PR #285, opened 2020-12-11
Improve testing infrastructure

ranocha:
I added a `Project.toml` for the tests, made CompatHelper track it, added doctests to CI, and introduced modules for tests to avoid cluttering the namespace.

### DIFF
--- a/docs/src/iterators.md
+++ b/docs/src/iterators.md
@@ -31,55 +31,44 @@ Now if you apply Jacobi iteration multiple times with the same matrix for just a
 ```jldoctest
 julia> using LinearAlgebra, SparseArrays, IterativeSolvers
 
-julia> A = sprand(10_000, 10_000, 10 / 10_000) + 20I;
+julia> A = spdiagm(-1 => -ones(3), 0 => 2*ones(4), 1 => -ones(3));
 
-julia> b1 = rand(10_000);
+julia> b1 = [1.0, 2, 3, 4];
 
-julia> b2 = rand(10_000);
+julia> b2 = [-1.0, 1, -1, 1];
 
-julia> x = rand(10_000);
+julia> x = [0.0, -1, 1, 0];
 
-julia> my_iterable = IterativeSolvers.jacobi_iterable(x, A, b1, maxiter = 4);
+julia> my_iterable = IterativeSolvers.jacobi_iterable(x, A, b1, maxiter = 2);
+
+julia> norm(b1 - A * x) / norm(b1)
+1.2909944487358056
 
 julia> for item in my_iterable
            println("Iteration for rhs 1")
        end
 Iteration for rhs 1
 Iteration for rhs 1
-Iteration for rhs 1
-Iteration for rhs 1
 
-julia> relative_residual = norm(b1 - A * x) / norm(b1);
+julia> norm(b1 - A * x) / norm(b1)
+0.8228507357554791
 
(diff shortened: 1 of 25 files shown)


docs/src/iterators.md
```diff
@@ -28,28 +28,42 @@ jacobi_iterable(x, A::SparseMatrixCSC, b; maxiter::Int = 10) =
...
+julia> for item in my_iterable
+           println("Iteration for rhs 2")
+       end
+Iteration for rhs 2
+Iteration for rhs 2
+Iteration for rhs 2
+Iteration for rhs 2
 
-@show norm(b2 - A * x) / norm(b2)
+julia> relative_residual ≈ norm(b2 - A * x) / norm(b2)
```
haampie, line 65: This doesn't really make sense, please remove the `relative_residual` variables and just show the values.

docs/src/iterators.md
```diff
@@ -28,28 +28,42 @@ jacobi_iterable(x, A::SparseMatrixCSC, b; maxiter::Int = 10) =
...
+false
 ```
 
 This would output:
```
haampie, line 69: This is duplicate output now, can you remove this?

docs/src/iterators.md
```diff
@@ -28,28 +28,42 @@ jacobi_iterable(x, A::SparseMatrixCSC, b; maxiter::Int = 10) =
 
 Now if you apply Jacobi iteration multiple times with the same matrix for just a few iterations, it makes sense to initialize the iterable only once and reuse it afterwards:
 
-```julia
-A = sprand(10_000, 10_000, 10 / 10_000) + 20I
-b1 = rand(10_000)
-b2 = rand(10_000)
-x = rand(10_000)
+```jldoctest
```
haampie, line 31: I'm not 100% sure whether it's a great idea to run a doctest with a problem size of 10_000. Also this is not really meant as a test, but rather as an example.

docs/src/iterators.md
ranocha: Done in 6fdc8f7.

docs/src/iterators.md
ranocha: Done in 6fdc8f7.

docs/src/iterators.md
ranocha: I modified the problem to be much smaller and suitable for CI doctests.

> I'm fine running it to ensure it doesn't break, but I don't think it should be treated as a test.

How would you do that?

haampie: Perfect! As is is good, it's basically just a smoke test (if it doesn't throw an error, it's fine).

haampie:
Can you "modularize" the new `deprecations.jl` tests too? Might have to rebase onto master.

And bump the version number?

ranocha:
> Can you "modularize" the new `deprecations.jl` tests too? Might have to rebase onto master.
> 
> And bump the version number?

Sure :+1:

ranocha:
Bump

mschauer:
I have a stupid question: is this actually testing the new setup? And we are dropping (test-) compatibility with Julia <1.2.

ViralBShah:
@haampie Good to merge?

haampie:
Yes. @mschauer I would be happy with only supporting julia 1.5.3+ because it has 5-arg mul, which is useful in some solvers.

PR #293, opened 2021-03-19
Make orthogonalization method chooseable for gmres

jkrch:
When using multithreading, Classical Gram-Schmidt with and without the DGKS correction step provides better runtimes then Modified Gram-Schmidt (since they use BLAS level 2 instead of level 1). 
The following shows GMRES solving the Bourchtein/atmosmodd linear system from SuiteSparse using the first from the two RHS which come with the matrix. All three runs converged after 1497 iterations (so at least for this linear system round off errors doesn't seem to be a problem).
orth_meth

Since CGS and DGKS are already implemented, I just made them choosable when calling gmres. Set MGS the default, since it might be the best without multithreading.

### DIFF
--- a/docs/src/linear_systems/gmres.md
+++ b/docs/src/linear_systems/gmres.md
@@ -13,7 +13,7 @@ gmres!
 
 The implementation pre-allocates a matrix $V$ of size `n` by `restart` whose columns form an orthonormal basis for the Krylov subspace. This allows BLAS2 operations when updating the solution vector $x$. The Hessenberg matrix is also pre-allocated.
 
-Modified Gram-Schmidt is used to orthogonalize the columns of $V$ by default, since it is more numerical stable then classical Gram-Schmidt. However will classical Gram-Schmidt (without or with the DGKS correction step) benefit more from multithreading.
+By default, modified Gram-Schmidt is used to orthogonalize the columns of $V$, since it is numerically more stable than classical Gram-Schmidt. Modified Gram-Schmidt is however inherently sequential, and if stability is not a concern, classical Gram-Schmidt can be used, which is implemented using BLAS2 operations. As a compromise the "DGKS criterion" can be used, which conditionally applies classical Gram-Schmidt repeatedly to stabilize it, and is typically one to two times slower than classical Gram-Schmidt. 
 
 The computation of the residual norm is implemented in a non-standard way, namely keeping track of a vector $\gamma$ in the null-space of $H_k^*$, which is the adjoint of the $(k + 1) \times k$ Hessenberg matrix $H_k$ at the $k$th iteration. Only when $x$ needs to be updated is the Hessenberg matrix mutated with Givens rotations.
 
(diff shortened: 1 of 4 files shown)


src/gmres.jl
```diff
@@ -184,15 +189,17 @@ function gmres!(x, A, b;
                 maxiter::Int = size(A, 2),
                 log::Bool = false,
                 initially_zero::Bool = false,
-                verbose::Bool = false)
+                verbose::Bool = false,
+                orth_meth = ModifiedGramSchmidt)
```
haampie, line 193: Needs to be documented

haampie:
```suggestion
                orth_meth::OrthogonalizationMethod = ModifiedGramSchmidt)
```

src/gmres.jl
```diff
@@ -109,7 +112,8 @@ function gmres_iterable!(x, A, b;
                          reltol::Real = sqrt(eps(real(eltype(b)))),
                          restart::Int = min(20, size(A, 2)),
                          maxiter::Int = size(A, 2),
-                         initially_zero::Bool = false)
+                         initially_zero::Bool = false,
+                         orth_meth = ModifiedGramSchmidt)
```
haampie, line 116:
```suggestion
                         orth_meth::OrthogonalizationMethod = ModifiedGramSchmidt)
```

haampie (changes requested):
I think modified Gramm-Schmidt is what people expect. I don't have the paper but I believe it should be dominated by rounding errors only exactly upon convergence, whereas classical GS can be dominated by noise before convergence.

haampie:
Can you include some comments explaining in not too technical terms what the user can choose and why he/she would want to do that? Basically comment on accuracy vs performance.

jkrch:
I can't get it running with `orth_meth::OrthogonalizationMethod = ModifiedGramSchmidt` :/
Thougth exporting `OrthogonalizationMethod` in `orthogonalize.jl` would help, but didn't...

haampie:
Oh, you need `::Type{<:OrthogonalizationMethod}` when passing the type along instead of an instance of it.

I'm also fine if you replace all occurances of

```orthogonalize_and_normalize!(..., ::Type{DGKS})` with ```orthogonalize_and_normalize!(..., ::DGKS)```

and make the user pass DGKS(), and the same for the other types

**Comment by jkrch** [comment]
Ah ok, gonna replace it

File: docs/src/linear_systems/gmres.md [hunk 3]
```diff
@@ -13,9 +13,9 @@ gmres!
 
 The implementation pre-allocates a matrix $V$ of size `n` by `restart` whose columns form an orthonormal basis for the Krylov subspace. This allows BLAS2 operations when updating the solution vector $x$. The Hessenberg matrix is also pre-allocated.
 
-Modified Gram-Schmidt is used to orthogonalize the columns of $V$.
+Modified Gram-Schmidt is used to orthogonalize the columns of $V$ by default, since it is more numerical stable then classical Gram-Schmidt. However will classical Gram-Schmidt (without or with the DGKS correction step) benefit more from multithreading.
```
haampie, line 16:
```suggestion
By default, modified Gram-Schmidt is used to orthogonalize the columns of $V$, since it is numerically more stable than classical Gram-Schmidt. Modified Gram-Schmidt is however inherently sequential, and if stability is not a concern, classical Gram-Schmidt can be used, which is implemented using BLAS2 operations. As a compromise the "DGKS criterion" can be used, which conditionally applies classical Gram-Schmidt repeatedly to stabilize it, and is typically one to two times slower than classical Gram-Schmidt. 
```

haampie (changes requested):

haampie:
Thanks for your contribution!

PR #297, opened 2021-05-14
Adding preconditioner to idrs

deltaeecs:
Here I added Pl preconditioner to idrs add tested it with IncompleteLU preconditioner.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "42fd0dbc-a981-5370-80f2-aaf504508153"
 version = "0.9.0"
 
 [deps]
+IncompleteLU = "40713840-3770-5561-ab4c-a76e7d0d7895"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,6 @@ uuid = "42fd0dbc-a981-5370-80f2-aaf504508153"
 version = "0.9.0"
 
 [deps]
-IncompleteLU = "40713840-3770-5561-ab4c-a76e7d0d7895"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/src/idrs.jl
+++ b/src/idrs.jl
@@ -24,6 +24,7 @@ shadow space.
 ## Keywords
 
 - `s::Integer = 8`: dimension of the shadow space;
+- `Pl::precT`: left preconditioner,
 - `abstol::Real = zero(real(eltype(b)))`,
   `reltol::Real = sqrt(eps(real(eltype(b))))`: absolute and relative
   tolerance for the stopping condition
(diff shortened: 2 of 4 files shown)


mschauer:
Nice! Did you follow literature or is it just clear how to add the preconditioner here? Pinging @astudillor, can you also have a cursory look?

src/idrs.jl
```diff
@@ -78,8 +80,8 @@ end
 end
 
 function idrs_method!(log::ConvergenceHistory, X, A, C::T,
-    s::Number, abstol::Real, reltol::Real, maxiter::Number; smoothing::Bool=false, verbose::Bool=false
-    ) where {T}
+    s::Number, Pl::precT, abstol::Real, reltol::Real, maxiter::Number; smoothing::Bool=false, verbose::Bool=false
+    ) where {T, precT}
```
mschauer, line 84: I think you never use `precT`, so you don't need that type parameter in the method signature

Project.toml
```diff
@@ -3,6 +3,7 @@ uuid = "42fd0dbc-a981-5370-80f2-aaf504508153"
 version = "0.9.0"
 
 [deps]
+IncompleteLU = "40713840-3770-5561-ab4c-a76e7d0d7895"
```
mschauer, line 6: We have a file for test dependencies, https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/master/test/Project.toml, we only need that dependency for the test, right?

src/idrs.jl
deltaeecs: Yes, I just followed the "gmres.jl" and added "precT". Maybe it will be useful in the future?

Project.toml
deltaeecs: I just checked the "gmres.jl" and found it use LU instead of the IncompleteLU I added in the dependency，shall I removed it?

mschauer: It makes sense to have a non-exact preconditioner in the test, or? Perhaps just move the 
```
IncompleteLU = "40713840-3770-5561-ab4c-a76e7d0d7895"
```
line to https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/master/test/Project.toml

deltaeecs:
> Nice! Did you follow literature or is it just clear how to add the preconditioner here? Pinging @astudillor, can you also have a cursory look?

hahaha
I followed the Matlab Code referred in the docs because I found idrs performs better  than gmres in my work~

Project.toml
deltaeecs: ok

mschauer:
Haha, you can also compare with https://github.com/astudillor/idrs/blob/master/idrs/idrs.py

test/idrs.jl
```diff
@@ -42,6 +43,25 @@ end
     @test norm(A * x - b) / norm(b) ≤ reltol
 end
 
+@testset "SparseMatrixCSC{$T, $Ti}" for T in (Float64, ComplexF64), Ti in (Int64, Int32)
```
mschauer, line 45:
```suggestion
@testset "SparseMatrixCSC{$T, $Ti} with preconditioner" for T in (Float64, ComplexF64), Ti in (Int64, Int32)
```

deltaeecs:
> Haha, you can also compare with https://github.com/astudillor/idrs/blob/master/idrs/idrs.py

Thank you for your help~
I think they are the same after checked the idrs.py~
I'm so happy since it's my first time to do a little work on git~

mschauer:
Great. Can I ask on what type of problem you are working such that IDR(s) beats GMRES?

deltaeecs:
> Great. Can I ask on what type of problem you are working such that IDR(s) beats GMRES?

It's a badly conditioned matrix equation on Computational Electromagneticcalled EFIE~

mschauer (approved):

deltaeecs:
> > Great. Can I ask on what type of problem you are working such that IDR(s) beats GMRES?
> 
> It's a badly conditioned matrix equation on Computational Electromagneticcalled EFIE~

Well it seems to be a problem of problem of setting on preconditioner. GMRES is still the best solver after I corrected it.

PR #304, opened 2021-10-30
use deterministic RNG initial state

### LINKED ISSUE #303: accept user-specified rng for random-number generation
Rather than calling `rand(...)`, it would be good to call `rand(rng, ...)` with an optionally user-specified `rng::AbstractRNG` keyword argument in algorithms that require pseudo-random initialization, e.g. lobpcg (https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/ae01dfe228138d0db9164eb343252150f6c0bfc6/src/lobpcg.jl#L872), bicgstabl (https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/ae01dfe228138d0db9164eb343252150f6c0bfc6/src/bicgstabl.jl#L38), and idrs (https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/ae01dfe228138d0db9164eb343252150f6c0bfc6/src/bicgstabl.jl#L38).

Rationale: [if deterministic results are desired](https://discourse.julialang.org/t/bicgstabl-from-iterativesolvers-jl-gives-different-result-every-time/70527) (as is often the case), it is better for the user to be able to specify an RNG than setting the *global* `Random.seed!`.

Moreover, arguably the *default* rng should be a new deterministic RNG object created with a fixed seed every time the iterative solver is initialized, so that the algorithm defaults to being deterministic.

(Note: "deterministic" ≠ "reproducible" … e.g. changing the Julia version could change the results.  But it is useful to be able to run the same code with the same software on the same hardware multiple times and get exactly the same results as much as possible.)

stevengj:
Closes #303.

Uses a deterministic RNG of `MersenneTwister(seed)` for generating initial vectors in iterative algorithms that use random initial vectors, controllable via an `rng` keyword.

(In some cases, the `rng` keyword is redundant, because the algorithm alternatively allows you to pass the initial vector directly, but it seemed better to me to have a consistent API that can be used with any of the iterative solvers doing pseudorandom initialization.)

(Bumped Julia requirement to 1.4 to support broadcasting RNG objects.)


mschauer (approved):

stevengj:
Weird, I can't replicate the test failure locally…

stevengj:
Okay, now I can replicate the problem.  It looks like the issue is the `lobpcg!` calls itself recursively, but I wasn't passing through the `rng` keyword, so it was generating the same pseudorandom numbers on each recursive call.

src/svdl.jl
```diff
@@ -176,7 +176,7 @@ end
 #########################
 
 function svdl_method!(log::ConvergenceHistory, A, l::Int=min(6, size(A,1)); k::Int=2l,
-    j::Int=l, rng::AbstractRNG=MersenneTwister(l), v0::AbstractVector = Vector{eltype(A)}(randn(rng, size(A, 2))) |> x->rmul!(x, inv(norm(x))),
+    j::Int=l, rng::AbstractRNG=MersenneTwister(seed), v0::AbstractVector = Vector{eltype(A)}(randn(rng, size(A, 2))) |> x->rmul!(x, inv(norm(x))),
```
mschauer, line 179: That looks better

mschauer:
The commit a4db20b had some build's passing, are the errors just from having different seeds and to tights bounds for stochastic tests?

stevengj:
> are the errors just from having different seeds and to tights bounds for stochastic tests?

Looks like it.   I switched to `master` and changed the `seed!` line in `test/lobpcg.jl` to `Random.seed!(Random.make_seed())`, to ensure a different seed on every run, and now I'm getting the same failures about 25% of the time — if I `include` that file 100 times, it fails 26 times.

Note that one of the failures is not just a tolerance thing, but a robustness problem: the Cholesky factorization [here](https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/6af7d5e5c7024c03158f65fd86d78ae6640c68bb/src/lobpcg.jl#L380) is sometimes failing because the matrix is indefinite (a nearly zero eigenvalue that is slightly indefinite … basically it looks like `X` is rank-deficient, in which case you should probably do some re-initialization?).

The other failure is on
```
test/lobpcg.jl:316
  Expression: all(isapprox.(adjoint(X) * (B * X), Matrix{T}(I, 3, 3), atol = 2 * n * tol))
```
where it looks like the tolerance might be too small.

(Not sure why you are using `all(isapprox.(..))` instead of just isapprox on the whole matrix here?)

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "IterativeSolvers"
 uuid = "42fd0dbc-a981-5370-80f2-aaf504508153"
-version = "0.9.1"
+version = "0.9.2"
```
mschauer, line 3:
```suggestion
version = "0.9.3"
```

mschauer:
Ping @haampie

mschauer:
Planning to merge this in some days.

mschauer:
@stevengj What to do with the tests? I agree that we can soften a bound for the second failure.

mschauer:
There is one test passing,
> CI / Julia 1.6 - ubuntu-latest - x64 - pull_request (pull_request) 
which seems to be that the lobpg issue is fixed on some level.  But shouldn't it pass on all systems as it is deterministic?

Issue #306, opened 2021-11-08
misleading documentation for minres


j-fu:
Hi, in the docstring for minres!, preconditioners are mentioned, however they are not in the signature of the methods.

PR #307, opened 2021-11-08
Docfix minres (#306)

mschauer:
Preconditions are mentioned, but not implemented. That's it, right? @j-fu.

Ref #306.


j-fu:
Yeah exactly, this was puzzling... Thanks for the fix.

haampie (approved):

PR #334, opened 2023-03-17
Modified CG to use axpy!/axpby! directly instead of broadcasting

vlc1:
When using IterativeSolvers.jl on custom vectors and matrices, I find it easier to add methods to generic BLAS routines (`scal!`, `mul!`, `axpy!`...) than customizing the broadcasting machinery on said types.

This is a very small modification that simply requires changing lines like
```julia
it.r .-= alpha .* it.c
```
to
```julia
axpy!(-alpha, it.c, it.r)
```
and so on as illustrated in the cg.jl file in this PR.

If this is a modification that is acceptable, I'll work my way through the other solvers (bicgstabl...).

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -48,15 +48,15 @@ function iterate(it::CGIterable, iteration::Int=start(it))
 
     # u := r + βu (almost an axpy)
     β = it.residual^2 / it.prev_residual^2
-    it.u .= it.r .+ β .* it.u
+    axpby!(true, it.r, β, it.u)
 
     # c = A * u
     mul!(it.c, it.A, it.u)
     α = it.residual^2 / dot(it.u, it.c)
 
     # Improve solution and residual
-    it.x .+= α .* it.u
-    it.r .-= α .* it.c
+    axpy!(α, it.u, it.x)
+    axpy!(-α, it.c, it.r)
 
     it.prev_residual = it.residual
     it.residual = norm(it.r)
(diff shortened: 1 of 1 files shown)


haampie:
I believe we've gone back and forth on this. Long ago there was overhead to broadcasting, so the BLAS routines were called. Then Julia improved and folks wanted broadcasting because it's more generic. I think I'm happy with the current broadcasting stuff, also because for example CG would need an `axpby` somewhere, which I think is not standard BLAS.

dkarrasch:
These functions are now owned by LinearAlgebra (and not `LinearAlgebra.BLAS`) and admit generic fallbacks, likely using broadcasting but I don't remember. Another option could be to introduce `_axp[b]y!`, that use broadcasting generically but may be overloaded with a custom `axpy!`-like implementation.

vlc1:
> These functions are now owned by LinearAlgebra (and not `LinearAlgebra.BLAS`) and admit generic fallbacks, likely using broadcasting but I don't remember. Another option could be to introduce `_axp[b]y!`, that use broadcasting generically but may be overloaded with a custom `axpy!`-like implementation.

Fair enough, I might have to dig into how to customize broadcasting for my own datatypes...

> These functions are now owned by LinearAlgebra (and not `LinearAlgebra.BLAS`) and admit generic fallbacks, likely using broadcasting but I don't remember. Another option could be to introduce `_axp[b]y!`, that use broadcasting generically but may be overloaded with a custom `axpy!`-like implementation.

That's pretty much what [Krylov.jl](https://github.com/JuliaSmoothOptimizers/Krylov.jl) does with some macros (`@kaxpy`...).

PR #336, opened 2023-03-28
fix typos

spaette:
docs/src/eigenproblems/power_method.md
docs/src/linear_systems/bicgstabl.md
src/history.jl

I used CamelCase in the following comment.

```
$ ed -s IterativeSolvers.jl/src/history.jl <<<'175,176p'
# If PartialHistory, there's no need to store a vector or matrix, instead
# store nothing or store a vector respectively.
$ 
```

### DIFF
--- a/docs/src/eigenproblems/power_method.md
+++ b/docs/src/eigenproblems/power_method.md
@@ -12,4 +12,4 @@ invpowm!
 ```
 
 ## Implementation details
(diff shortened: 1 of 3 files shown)


spaette:
> I don't know whether partialhistory might be CamelCase.

@jiahao 

Is there any interest in this pull request?

spaette:
> I don't know whether partialhistory might be CamelCase.

@mschauer

Should the comment be left as is or should CamelCase be used?

Issue #325, opened 2022-09-01
Iterator protocol for IDR(s)


mschauer:
I would love to see that IDR(s) joins the iterator-interface like the other solvers. That's also important for SciML interoperability where IDR(s) is missing from http://linearsolve.sciml.ai/dev/solvers/solvers/#IterativeSolvers.jl .

PR #345, opened 2023-08-03
IDR(s) as iterator

mschauer:
Addresses #325


mschauer:
I'll merge if there are no further comments.

Issue #358, opened 2024-02-12
Does not compile for Julia <1.5


kylebeggs:
The keyword arg syntax [here](https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/077214fe13e692358395a81ddd692e64056a4b2c/src/idrs.jl#L149) breaks compilation on Julia 1.3,1.4.

kylebeggs:
```julia
julia> using IterativeSolvers
[ Info: Precompiling IterativeSolvers [42fd0dbc-a981-5370-80f2-aaf504508153]
ERROR: LoadError: LoadError: syntax: invalid keyword argument syntax "smoothing"
Stacktrace:
 [1] top-level scope at /home/kylebeggs/.julia/packages/IterativeSolvers/QnUAe/src/idrs.jl:148
 [2] include(::Module, ::String) at ./Base.jl:377
 [3] include(::String) at /home/kylebeggs/.julia/packages/IterativeSolvers/QnUAe/src/IterativeSolvers.jl:7
 [4] top-level scope at /home/kylebeggs/.julia/packages/IterativeSolvers/QnUAe/src/IterativeSolvers.jl:24
 [5] include(::Module, ::String) at ./Base.jl:377
 [6] top-level scope at none:2
 [7] eval at ./boot.jl:331 [inlined]
 [8] eval(::Expr) at ./client.jl:449
 [9] top-level scope at ./none:3
in expression starting at /home/kylebeggs/.julia/packages/IterativeSolvers/QnUAe/src/idrs.jl:148
in expression starting at /home/kylebeggs/.julia/packages/IterativeSolvers/QnUAe/src/IterativeSolvers.jl:24
ERROR: Failed to precompile IterativeSolvers [42fd0dbc-a981-5370-80f2-aaf504508153] to /home/kylebeggs/.julia/compiled/v1.4/IterativeSolvers/ef2NV_qUW8c.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1272
 [3] _require(::Base.PkgId) at ./loading.jl:1029
 [4] require(::Base.PkgId) at ./loading.jl:927
 [5] require(::Module, ::Symbol) at ./loading.jl:922

julia> versioninfo()
Julia Version 1.4.2
Commit 44fa15b150* (2020-05-23 18:35 UTC)
Platform Info:
[...]

PR #359, opened 2024-02-12
fix kwargs in idrs_iterable! call in idrs_method!

kylebeggs:
Addresses #358


mschauer (approved):

kylebeggs:
@mschauer Can we merge?

PR #362, opened 2024-05-16
Modification to prevent scalar indexing on CUDA devices

oscarvanderheide:
Scalar indexing on CUDA devices results in poor performance. This modification uses dot syntax to prevent scalar indexing.


ViralBShah (approved):

oscarvanderheide:
Would it be possible to have this change merged now that it is approved? I see a check related to documentation is unsuccessful but I don't think it it is related to the change in this pull request?

